PR #8, opened 2021-10-29
add a few tests

mrr00b00t:
Try to build some tests for Parallel and Sequential


test/transforms.jl
```diff
@@ -223,10 +223,47 @@
   end
 
   @testset "Sequential" begin
-    # TODO
+    x = rand(Normal(0,10), 1500)
+    y = x + rand(Normal(0,2), 1500)
+    z = y + rand(Normal(0,5), 1500)
+    t = Table(; x, y, z)
+    S = Scale(low=0.2, high=0.8) → EigenAnalysis(:VDV)
```
juliohm, line 230: Please use T for transforms, there are too many letters already.

PR #12, opened 2021-11-01
solve unecessary use of inv function

### LINKED ISSUE #10: Avoid inverting matrices in EigenAnalysis
We can be more clever and avoid inversions of S here:

https://github.com/JuliaML/TableTransforms.jl/blob/1841f05f9b8192388fcb2960daee00a9eef1aece/src/transforms/eigenanalysis.jl#L38-L48

We should be able to express these inverses by swapping the order of the products and by transposing the matrices instead.

mrr00b00t:
Try to fix #10.
All tests passing locally (not considering that image test that passes/don't aleatory)

### DIFF
--- a/src/transforms/eigenanalysis.jl
+++ b/src/transforms/eigenanalysis.jl
@@ -36,21 +36,17 @@ function pcaproj(λ, V)
 end
 
 function drsproj(λ, V)
-  λₛ = sqrt.(λ)
-  Λᵢ = Diagonal(1 ./ λₛ)
-  Λₛ = Diagonal(λₛ)
-  S = V * Λᵢ
-  Sᵢ = Λₛ * transpose(V)
-  S, Sᵢ
+  Λ = Diagonal(sqrt.(λ))
+  S = V * inv(Λ)
+  S⁻¹ = Λ * transpose(V)
+  S, S⁻¹
 end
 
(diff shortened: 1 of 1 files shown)


src/transforms/eigenanalysis.jl
```diff
@@ -36,15 +36,21 @@ function pcaproj(λ, V)
 end
 
 function drsproj(λ, V)
-  Λ = Diagonal(sqrt.(λ))
-  S = V * inv(Λ)
-  S, inv(S)
+  λₛ = sqrt.(λ)
+  Λᵢ = Diagonal(1 ./ λₛ)
```
juliohm, line 40: Be consistent. In other parts of the code you use `^-1` to represent inverse matrices, and here you use `i`.

src/transforms/eigenanalysis.jl
```diff
@@ -36,15 +36,21 @@ function pcaproj(λ, V)
...
+  Λₛ = Diagonal(λₛ)
```
juliohm, line 41: Can we just use `\Lambda = Diagonal(sqrt.(\lambda))` and the corresponding inverse?

juliohm (changes requested):
Lacking consistency.

juliohm (approved):

PR #15, opened 2021-11-14
wip: implement Rename transform

Omar-Elrefaei:
This should be fully functional, but I think the usage of `push!!` will need to be redesigned to be more functional and efficient like other transforms. That is the place where I think I need a little help: I am not sure how to make it so that I don't need `push!!` and use `𝒯 = (; zip(anames, acols)...)` instead like the rest of the of the transforms.

### DIFF
--- a/src/TableTransforms.jl
+++ b/src/TableTransforms.jl
@@ -7,7 +7,7 @@ module TableTransforms
 using Tables
 using ScientificTypes
 using Distributions: Normal
-using Transducers: tcollect, push!!
+using Transducers: tcollect
 using LinearAlgebra
 using Statistics
 
@@ -29,10 +29,10 @@ export
   colapply, colrevert,
 
   # built-in
-  Identity,
   Select,
   Reject,
-  Rename3,
+  Rename,
+  Identity,
   Center,
   Scale,
   MinMax,
--- a/src/TableTransforms.jl
+++ b/src/TableTransforms.jl
@@ -7,7 +7,7 @@ module TableTransforms
 using Tables
 using ScientificTypes
 using Distributions: Normal
-using Transducers: tcollect
+using Transducers: tcollect, push!!
 using LinearAlgebra
 using Statistics
 
@@ -31,6 +31,7 @@ export
   # built-in
   Select,
   Reject,
+  Rename,
   Identity,
   Center,
   Scale,
--- a/src/transforms/rename.jl
+++ b/src/transforms/rename.jl
@@ -31,9 +31,7 @@ function _rename(names, table)
   oldnames = Tables.columnnames(table)
 
   # check if requested renames exist in the table
-  diff = setdiff(keys(names), oldnames) .|> String |> Tuple
-  @assert length(diff) == 0
-    "The following column[s] were not found in the source table: $diff"
+  @assert keys(names) ⊆ oldnames "invalid column names"
 
   newnames = map(oldnames) do oldname
     oldname in keys(names) ? names[oldname] : oldname
--- a/src/transforms/rename.jl
+++ b/src/transforms/rename.jl
@@ -3,9 +3,9 @@
 # ------------------------------------------------------------------
 
 """ 
-  Rename(:col₁ => :newcol₁, :col₂ => :newcol₂, ..., :col₁ => :newcolₙ))
+  Rename(:col₁ => :newcol₁, :col₂ => :newcol₂, ..., :col₁ => :newcolₙ)
 
-Tha transform that renames `col₁` to `newcol₁`, `col₂` to `newcol₂`, ...
+The transform that renames `col₁` to `newcol₁`, `col₂` to `newcol₂`, ...
 """
 struct Rename <: Stateless
   names::Dict{Symbol,Symbol}
@@ -21,12 +21,9 @@ function apply(transform::Rename, table)
     end
   end
   acols = [i for i in Tables.columns(table)]
-  𝒯 = (; zip(newnames, acols)...)
-  𝒯 |> Tables.materializer(table)
+  𝒯 = (; zip(newnames, acols)...) |> Tables.materializer(table)
   𝒯, nothing
 end
-
-
 
 function revert(transform::Rename, table)
   # reversing the key-value pairs of the Dict
--- a/src/transforms/rename.jl
+++ b/src/transforms/rename.jl
@@ -3,7 +3,7 @@
 # ------------------------------------------------------------------
 
 """
-  Rename(:col₁ => :newcol₁, :col₂ => :newcol₂, ..., :col₁ => :newcolₙ)
+    Rename(:col₁ => :newcol₁, :col₂ => :newcol₂, ..., :col₁ => :newcolₙ)
 
 The transform that renames `col₁` to `newcol₁`, `col₂` to `newcol₂`, ...
 """
--- a/src/transforms/rename.jl
+++ b/src/transforms/rename.jl
@@ -2,7 +2,7 @@
 # Licensed under the MIT License. See LICENSE in the project root.
 # ------------------------------------------------------------------
 
-""" 
+"""
   Rename(:col₁ => :newcol₁, :col₂ => :newcol₂, ..., :col₁ => :newcolₙ)
 
 The transform that renames `col₁` to `newcol₁`, `col₂` to `newcol₂`, ...
@@ -14,10 +14,10 @@ end
 function apply(transform::Rename, table)
   oldnames = Tables.columnnames(table)
   newnames = map(oldnames) do oldname
-    if oldname in keys(transform.names)
-      return transform.names[oldname]
+    if oldname in keys(transform.names) 
+      transform.names[oldname]
     else
-      return oldname
+      oldname
     end
   end
   acols = [i for i in Tables.columns(table)]
--- a/src/transforms/rename.jl
+++ b/src/transforms/rename.jl
@@ -13,7 +13,7 @@ end
 
 function Rename(names...)
   sympair(x) = Symbol(first(x)) => Symbol(last(x))
-  map(sympair , names) |> Dict |> Rename
+  map(sympair, names) |> Dict |> Rename
 end
 
 function apply(transform::Rename, table)
@@ -33,6 +33,10 @@ end
 
 function _rename(names, table)
   oldnames = Tables.columnnames(table)
+  @show  oldnames, names
+
+  dif = setdiff(keys(names), oldnames) |> Tuple
+  @assert length(dif) == 0 "The following column[s] were not found in the source table $dif"
 
   newnames = map(oldnames) do oldname
     oldname in keys(names) ? names[oldname] : oldname
--- a/src/transforms/rename.jl
+++ b/src/transforms/rename.jl
@@ -33,6 +33,7 @@ function _rename(names, table)
   # check if requested renames exist in the table
   @assert keys(names) ⊆ oldnames "invalid column names"
 
+  # use new names if necessary
   newnames = map(oldnames) do oldname
     oldname in keys(names) ? names[oldname] : oldname
   end
--- a/src/transforms/rename.jl
+++ b/src/transforms/rename.jl
@@ -1,0 +1,40 @@
+# ------------------------------------------------------------------
+# Licensed under the MIT License. See LICENSE in the project root.
+# ------------------------------------------------------------------
+
+""" 
+  Rename(:col₁ => :newcol₁, :col₂ => :newcol₂, ..., :col₁ => :newcolₙ))
+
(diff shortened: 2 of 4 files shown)


src/TableTransforms.jl
```diff
@@ -7,7 +7,7 @@ module TableTransforms
 using Tables
 using ScientificTypes
 using Distributions: Normal
-using Transducers: tcollect
+using Transducers: tcollect, push!!
```
juliohm, line 10: We don't need `push!!` to implement `Rename`.

src/TableTransforms.jl
```diff
@@ -29,9 +29,10 @@ export
   colapply, colrevert,
 
   # built-in
+  Identity,
   Select,
   Reject,
-  Identity,
+  Rename3,
```
juliohm, line 35: This export is not correct. Also keep `Identity` in the line it was placed.

src/transforms/rename.jl
```diff
@@ -0,0 +1,43 @@
+# ------------------------------------------------------------------
+# Licensed under the MIT License. See LICENSE in the project root.
+# ------------------------------------------------------------------
+
+""" 
+    Rename(Dict(:col₁ => :newcol₁, :col₂ => :newcol₂, ..., :col₁ => :newcolₙ))
```
juliohm, line 6: This is an internal signature. After we make it work with `Dict` we will only document the VarArg version to end users. So this string should read `Rename(:col1 => :newcol1, :col2 => :newcol2, ...)`.

src/transforms/rename.jl
```diff
@@ -0,0 +1,43 @@
...
+
+Tha transform that renames `col₁` to `newcol₁`, `col₂` to `newcol₂`, ...
+""" 
+struct Rename3 <: Stateless
```
juliohm, line 10: We don't want `Rename3`.

src/transforms/rename.jl
```diff
@@ -0,0 +1,43 @@
...
+    names::Dict{Symbol,Symbol}
+end
+
+function apply(transform::Rename3, table)
+    new_table = (;)
```
juliohm, line 15: Before we start applying the transform, we need to make sure that the input table has the columns specified in `Rename`.

src/transforms/rename.jl
```diff
@@ -0,0 +1,43 @@
...
+
+    for col_name in Tables.columnnames(table)
+        col_value = Tables.getcolumn(table, col_name)
```
juliohm, line 18: Try to follow the code style in the other transforms. We avoid variable names with underscore for example. We use 2 whitespaces for indentation, etc.

juliohm (changes requested):
Attached a first round of reviews.

Omar-Elrefaei:
Ooof, my bad for half of these things. I guess that what you get from sending a PR of after-midnight code without a review. I will fix the formatting and naming right away, but that I am not 100% sure about is how to phase out `push!!` ?

Omar-Elrefaei:
Is it as simple as looping through the results of `Tables.columnnames(table)`, replacing the names that need to be renamed, and then `(; zip(names, cols)...)` ??

juliohm:
Exactly, you can write it in a single:

```julia
newcols = map(cols) do col
  # do the renaming
end
```

See the other transforms for examples.

src/TableTransforms.jl
Omar-Elrefaei: > Also keep Identity in the line it was placed.
I was making it inline with the order of includes in `transforms.jl`

Omar-Elrefaei:
I'm sorry, I screwed up git a bit. I'll comment when it's ready for review.

src/transforms/rename.jl
```diff
@@ -0,0 +1,37 @@
...
+end
+
+function apply(transform::Rename, table)
+  oldnames = Tables.columnnames(table)
+  newnames = map(oldnames) do oldname
+    if oldname in keys(transform.names) 
+      transform.names[oldname]
+    else
+      oldname
+    end
```
Omar-Elrefaei, line 21: This could be a simple:
```(oldname in keys(transform.names))  ? transform.names[oldname] : oldname```
or even:
`if (oldname in keys(transform.names)) transform.names[oldname] else oldname end`

**Comment by Omar-Elrefaei** [comment]
I believe its ready for another review.

File: src/transforms/rename.jl [hunk 8]
```diff
@@ -0,0 +1,37 @@
...
+"""
+  Rename(:col₁ => :newcol₁, :col₂ => :newcol₂, ..., :col₁ => :newcolₙ)
+
+The transform that renames `col₁` to `newcol₁`, `col₂` to `newcol₂`, ...
+"""
+struct Rename <: Stateless
+  names::Dict{Symbol,Symbol}
+end
+
+function apply(transform::Rename, table)
```
juliohm, line 14: I think we need to `@assert` to make sure that the dictionary doesn't contain invalid oldnames. That is, before the map, we should make sure that the `keys(transform.names) \subseteq oldnames`.

src/transforms/rename.jl
juliohm: The ternary operator is nice, specially if you save a new variable `names = transform.names` before the map to shorten it up.

src/transforms/rename.jl
```diff
@@ -0,0 +1,37 @@
...
+  oldnames = Tables.columnnames(table)
+  newnames = map(oldnames) do oldname
+    if oldname in keys(transform.names) 
+      transform.names[oldname]
+    else
+      oldname
+    end
+  end
+  acols = [i for i in Tables.columns(table)]
```
juliohm, line 23: I think this is not fully compatible with the Tables.jl interface. The result of `Tables.column` is not necessarily iterable. We need to use `Tables.getcolumn(cols, name)` to access a specific column.

src/transforms/rename.jl
```diff
@@ -0,0 +1,37 @@
...
+  𝒯 = (; zip(newnames, acols)...) |> Tables.materializer(table)
+  𝒯, nothing
+end
+
+function revert(transform::Rename, table)
+  # reversing the key-value pairs of the Dict
+  new_names = Dict()
+  for (new, old) in transform.names
```
juliohm, line 31: I think you you misplaced `new` and `old` in this loop. Can you please add tests to the transform so that we catch basic bugs?

src/transforms/rename.jl
```diff
@@ -0,0 +1,37 @@
...
+    new_names[old] = new
+  end
+
+  reversed_transform = Rename(new_names)
+  apply(reversed_transform, table) |> first
```
juliohm, line 36: We don't want create a new transform to revert. What we need to do is save the old names in the cache and use the cache to revert the operation.

juliohm (changes requested):
Review attached.

src/transforms/rename.jl
juliohm: Alternatively you can refactor this common set of instructions into an separate function `_rename` that does the actual work, and use this function in multiple places. This will facilitate future rearrangements of the code, and is good software engineering practice.

src/transforms/rename.jl
Omar-Elrefaei: I will add tests (might take a little of poking around given this is my first forte into package testing.)
But I have been testing continuously while developing (I don't feel comfortable while writing code otherwise) and it works perfectly. Both reverting and appling invalid oldnames are handled as expected.
I can see why you said they might be misplaced, semantically I should probably swap them both in the for loop and the dict assignment (line32).

src/transforms/rename.jl
Omar-Elrefaei: I can add an assert if you want (maybe it's a UX issue?), but right now due to the `if oldname in keys(transform.names)` invalid oldnames are gracefully ignored.

src/transforms/rename.jl
Omar-Elrefaei: TIL. I noticed that it was not working with DataFrames, but it was also not working with JuliaDB, and I saw other `Tables.column` instances throughout the codebase so I thought it was fine for now. Turns out JuliaDB does not fully implement Tables.jl, and your other usages were not iterating.

src/transforms/rename.jl
juliohm: Yes, we want to assert to alert end users that they specified incorrect column names 👍🏽

src/transforms/rename.jl
juliohm: Are you familiar with Julia's testing infrastructure? All you need to do is activate the folder of the package and type `] test`. You then add more tests and rerun.

Omar-Elrefaei: > Are you familiar with Julia's testing infrastructure?

No

> All you need to do is activate the folder of the package and type `] test`

that sounds like a simple workflow. I have used the package manager a lot but as a user, never as a package dev.
I can see the simple format in `test/transforms.jl` and I can add to that. 
Question: is there a way to just run the tests I'm developing instead of everything in there?

juliohm: Never used but I think Retest.jl has this functionality: https://github.com/JuliaTesting/ReTest.jl

juliohm:
@Omar-Elrefaei did you have time to update this PR?

Omar-Elrefaei:
Unfortunately no, I haven't been getting a chance to continue the work.

juliohm:
@Omar-Elrefaei you mean that this PR should be closed?

Omar-Elrefaei:
> @Omar-Elrefaei you mean that this PR should be closed?

No, if not today after my dayjob, I'll definitely be sending updates in the weekend.

src/transforms/rename.jl
Omar-Elrefaei: I am gonna do it this second way, because I am not sure if there is anything useful to put in that cache.

So it is ok to: 
- refactor that `_rename`, 
- call it twice instead of creating another transform, 
- and not use `cache`. (pass it around as `nothing` instead)

Omar-Elrefaei:
I think it's ready for another review.
I have:
- [x] added the assert
- [x] added tests
- [x] refactored `_rename` out
- [x] made the varargs constructor

src/transforms/rename.jl
```diff
@@ -0,0 +1,50 @@
...
+function apply(transform::Rename, table)
+  _rename(transform.names, table)
+end
+
+function revert(transform::Rename, table, cache)
+  # reversing the key-value pairs of the Dict
+  newnames = Dict()
+  for (old, new) in transform.names
+    newnames[new] = old
+  end
```
juliohm, line 28:
```suggestion
  newnames = Dict(new => old for (old, new) in transform.names)
```

src/transforms/rename.jl
```diff
@@ -0,0 +1,47 @@
...
+end
+
+
+function _rename(names, table)
+  oldnames = Tables.columnnames(table)
+
+  # check if requested renames exist in the table
+  diff = setdiff(keys(names), oldnames) .|> String |> Tuple
+  @assert length(diff) == 0
+    "The following column[s] were not found in the source table: $diff"
```
juliohm, line 36:
```suggestion
  @assert keys(names) ⊆ oldnames "invalid column names"
```

src/transforms/rename.jl
```diff
@@ -0,0 +1,45 @@
...
+end
+
+
+function _rename(names, table)
+  oldnames = Tables.columnnames(table)
+
+  # check if requested renames exist in the table
+  @assert keys(names) ⊆ oldnames "invalid column names"
+
+  newnames = map(oldnames) do oldname
```
juliohm, line 36:
```suggestion
  # use new names if necessary
  newnames = map(oldnames) do oldname
```

juliohm (approved):

src/transforms/rename.jl
```diff
@@ -0,0 +1,46 @@
+# ------------------------------------------------------------------
+# Licensed under the MIT License. See LICENSE in the project root.
+# ------------------------------------------------------------------
+
+"""
+  Rename(:col₁ => :newcol₁, :col₂ => :newcol₂, ..., :col₁ => :newcolₙ)
```
juliohm, line 6:
```suggestion
    Rename(:col₁ => :newcol₁, :col₂ => :newcol₂, ..., :col₁ => :newcolₙ)
```

juliohm:
Awesome improvements @Omar-Elrefaei , I am merging it.

The test that is failing is unrelated. For some reason one of our visual tests is failing randomly. Do you think you can take a look at this issue? Maybe we are missing some `Random.seed!` call somewhere?

Omar-Elrefaei:
Great! 
Thank you very much @juliohm for the patient and valuable feedback.

I'll open an issue so that not to forget to investigate the failing tests. 
Looking forward to my next PR 😄

PR #23, opened 2022-01-28
New Select and Reject

### LINKED ISSUE #14: Allow regular expression in Select/Reject
We should be able to write `Select(r"foo")` to select all columns containing the string "foo".

eliascarv:
Update Select and Reject to add Regex support.

Fix #14

### DIFF
--- a/src/transforms/select.jl
+++ b/src/transforms/select.jl
@@ -9,7 +9,6 @@ const ColSelector = Union{Vector{Symbol}, Regex}
     Select([col₁, col₂, ..., colₙ])
     Select((col₁, col₂, ..., colₙ))
     
-
 The transform that selects columns `col₁`, `col₂`, ..., `colₙ`.
     
     Select(Regex)
@@ -27,6 +26,8 @@ Select(cols::NTuple{N, T}) where {N, T<:Union{AbstractString, Symbol}} =
 
 Select(cols::Vector{T}) where {T<:AbstractString} =
   Select(Symbol.(cols))
+
+Base.:(==)(a::Select, b::Select) = a.cols == b.cols
 
 isrevertible(::Type{<:Select}) = true
 
@@ -89,26 +90,34 @@ end
 """
     Reject(col₁, col₂, ..., colₙ)
     Reject([col₁, col₂, ..., colₙ])
+    Reject((col₁, col₂, ..., colₙ))
 
 The transform that discards columns `col₁`, `col₂`, ..., `colₙ`.
+
+    Reject(Regex)
+Discards the columns that match with Regex.
 """
-struct Reject{N} <: Stateless
-  cols::NTuple{N,Symbol}
+struct Reject{S<:ColSelector} <: Stateless
+  cols::S
 end
 
-Reject(cols::NTuple{N,AbstractString}) where {N} =
+Reject(cols::T...) where {T<:Union{AbstractString, Symbol}} = 
+  Reject(cols)
+
+Reject(cols::NTuple{N, T}) where {N, T<:Union{AbstractString, Symbol}} =
+  Reject(collect(cols))
+
+Reject(cols::Vector{T}) where {T<:AbstractString} =
   Reject(Symbol.(cols))
 
-Reject(cols::AbstractVector) = Reject(Tuple(cols))
-
-Reject(cols...) = Reject(cols)
+Base.:(==)(a::Reject, b::Reject) = a.cols == b.cols
 
 isrevertible(::Type{<:Reject}) = true
 
 function apply(transform::Reject, table)
   allcols = Tables.columnnames(table)
-  reject  = collect(transform.cols)
-  select  = Tuple(setdiff(allcols, reject))
+  reject  = _selectedcols(transform.cols, allcols)
+  select  = setdiff(allcols, reject)
   strans  = Select(select)
   newtable, scache = apply(strans, table)
   newtable, (strans, scache)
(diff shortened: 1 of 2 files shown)


eliascarv:
This PR is ready for review @juliohm.

test/transforms.jl
```diff
@@ -10,6 +10,12 @@
     f = rand(4000)
     t = Table(; a, b, c, d, e, f)
 
+    x1 = rand(4000)
+    x2 = rand(4000)
+    y1 = rand(4000)
+    y2 = rand(4000)
+    t₂ = Table(; x1, x2, y1, y2)
+
```
juliohm, line 18: Let's move this to right before the corresponding test lines. We should probably stick to the surrounding convention and use t for the variable name. Every test set overwrites a new t and runs a couple of tests.

test/transforms.jl
```diff
@@ -55,10 +61,45 @@
...
+    T = Select(["d", "c", "b"])
+    n, c = apply(T, t)
+    @test Tables.columnnames(n) == (:d, :c, :b)
+    tₒ = revert(T, n, c)
+    @test t == tₒ
+
+    # selection with Regex
+    T = Select(r"[dcb]")
+    n, c = apply(T, t)
+    @test Tables.columnnames(n) == (:b, :c, :d) # a ordem das colunas é a mesma da table original - obs: traduzir
```
juliohm, line 81: Please use English always, even in comments.

test/transforms.jl
```diff
@@ -55,10 +61,45 @@
...
+    tₒ = revert(T, n, c)
+    @test t₂ == tₒ
+
     # selection with single column
-    @test (Select(:a) == Select("a") ==
-           Select((:a,)) == Select(("a",)) ==
-           Select([:a]) == Select(["a"]))
+    @test (Select(:a).cols == Select("a").cols ==
+           Select((:a,)).cols == Select(("a",)).cols ==
+           Select([:a]).cols == Select(["a"]).cols)
```
juliohm, line 102: Try to rewrite these tests without accessing internal fields. Internal fields are not part of the public API, and may change in future refactoring of the code.

juliohm (changes requested):
Awesome progress. Please take a look at the attached suggestions.

Also, we should make sure that the Reject transform is also working with regex.

test/transforms.jl
eliascarv: OK, done!

test/transforms.jl
eliascarv: Done.

test/transforms.jl
eliascarv: The comparison operator has been overloaded to fix this issue.
Done!

src/transforms/select.jl
```diff
@@ -2,29 +2,43 @@
...
 """
     Select(col₁, col₂, ..., colₙ)
     Select([col₁, col₂, ..., colₙ])
-
+    Select((col₁, col₂, ..., colₙ))
+    
 The transform that selects columns `col₁`, `col₂`, ..., `colₙ`.
+    
+    Select(Regex)
+Selects the columns that match with Regex.
```
juliohm, line 15:
```suggestion
    Select(regex)

Selects the columns that match with `regex`.
```

src/transforms/select.jl
```diff
@@ -2,29 +2,43 @@
...
-Select(cols...) = Select(cols)
+Select(cols::Vector{T}) where {T<:AbstractString} =
+  Select(Symbol.(cols))
+
+Base.:(==)(a::Select, b::Select) = a.cols == b.cols
 
 isrevertible(::Type{<:Select}) = true
 
+_selectedcols(cols::Vector{Symbol}, allcols) = cols
+_selectedcols(cols::Regex, allcols) =
```
juliohm, line 35: What about renaming this function to the verb `_select`?

src/transforms/select.jl
```diff
@@ -76,26 +90,34 @@ end
 """
     Reject(col₁, col₂, ..., colₙ)
     Reject([col₁, col₂, ..., colₙ])
+    Reject((col₁, col₂, ..., colₙ))
 
 The transform that discards columns `col₁`, `col₂`, ..., `colₙ`.
+
+    Reject(Regex)
+Discards the columns that match with Regex.
```
juliohm, line 98:
```suggestion
    Reject(regex)

Discards the columns that match with `regex`.
```

src/transforms/select.jl
```diff
@@ -2,29 +2,43 @@
 # Licensed under the MIT License. See LICENSE in the project root.
 # ------------------------------------------------------------------
 
+const ColSelector = Union{Vector{Symbol}, Regex}
```
juliohm, line 5: What about renaming this to `ColSpec`? It solves the Select/Reject nomenclature issue.

juliohm (changes requested):
Great PR. Just a few more changes before we can go ahead and merge.

src/transforms/select.jl
eliascarv: I think that's a good name.
I will rename `ColSelector` to `ColSpec`.

src/transforms/select.jl
eliascarv: Done!

src/transforms/select.jl
eliascarv: Done.

src/transforms/select.jl
eliascarv: Done!

juliohm (approved):

juliohm:
Great contribution 👍🏽

PR #24, opened 2022-01-31
TableSelection

### LINKED ISSUE #16: Lazy Select/Reject
Currently our `Select/Reject` are greedy and generate copies of the input tables. We could create a lazy type for selections similar to what TableOperations.jl does.

eliascarv:
This PR implements a lazy type for selections/rejections.

Fix #16

### DIFF
--- a/src/transforms/select.jl
+++ b/src/transforms/select.jl
@@ -2,6 +2,32 @@
 # Licensed under the MIT License. See LICENSE in the project root.
 # ------------------------------------------------------------------
 
+struct TableSelection{T} 
+  table::T
+  cols::Vector{Symbol}
+end
+
+function Base.getproperty(ts::TableSelection, col::Symbol)
+  if col ∈ ts.cols
+    Tables.getcolumn(ts.table, col)
+  else
+    error("This table has no column $(col).")
+  end
+end
+
+Base.getindex(ts::TableSelection, row::Int, col::Symbol) =
+  getproperty(ts, col)[row]
+
+Base.getindex(ts::TableSelection, row::Int, col::Int) =
+  getproperty(ts, ts.cols[col])[row]
+
+Tables.istable(::Type{<:TableSelection}) = true
+Tables.columnaccess(::Type{<:TableSelection}) = true
+Tables.columns(ts::Type{<:TableSelection}) = ts
+Tables.columnnames(ts::TableSelection) = ts.cols
+Tables.getcolumn(ts::TableSelection, col::Symbol) = getproperty(ts, col)
+Tables.getcolumn(ts::TableSelection, col::Int) = getproperty(ts, ts.cols[col])
+
 const ColSpec = Union{Vector{Symbol}, Regex}
 
 """
--- a/src/transforms/select.jl
+++ b/src/transforms/select.jl
@@ -24,8 +24,13 @@ function Tables.schema(ts::TableSelection)
   return Tables.Schema(ts.cols, selecttypes)
 end
 
+Tables.materializer(::Type{TableSelection{T}}) where {T} = 
+  Tables.materializer(T)
+
+Base.:(==)(a::TableSelection, b::TableSelection) = a.table == b.table
+
 function Base.show(io::IO, ts::TableSelection{T}) where {T}
-  println(io, "TableSelection of $(T)")
+  println(io, "TableSelection")
   pretty_table(io, ts, vcrop_mode=:middle)
 end
 
@@ -89,11 +94,13 @@ function apply(transform::Select, table)
   rcols = [Tables.getcolumn(cols, name) for name in reject]
 
   # table with selected columns
-  𝒯 = (; zip(select, scols)...)
-  stable = 𝒯 |> Tables.materializer(table)
+  # 𝒯 = (; zip(select, scols)...)
+  # stable = 𝒯 |> Tables.materializer(table)
 
-  stable, (reject, rcols, sperm, rinds)
+  TableSelection(table, select), (reject, rcols, sperm, rinds)
 end
+
+revert(::Select, newtable::TableSelection, cache) = newtable.table
 
 function revert(::Select, newtable, cache)
   # selected columns
--- a/src/transforms/select.jl
+++ b/src/transforms/select.jl
@@ -11,8 +11,8 @@ Tables.istable(::Type{<:TableSelection}) = true
 Tables.columnaccess(::Type{<:TableSelection}) = true
 Tables.columns(ts::Type{<:TableSelection}) = ts
 Tables.columnnames(ts::TableSelection) = ts.cols
-Tables.getcolumn(ts::TableSelection, col::Symbol) = getproperty(ts, col)
-Tables.getcolumn(ts::TableSelection, col::Int) = getproperty(ts, ts.cols[col])
+Tables.getcolumn(ts::TableSelection, col::Union{Symbol, Int}) = 
+  Tables.getcolumn(ts.table, col)
 
 const ColSpec = Union{Vector{Symbol}, Regex}
 
--- a/src/transforms/select.jl
+++ b/src/transforms/select.jl
@@ -7,14 +7,28 @@ struct TableSelection{T}
   cols::Vector{Symbol}
 end
 
+function Base.:(==)(a::TableSelection, b::TableSelection)
+  if a.cols == b.cols
+    all(Tables.getcolumn(a, col) == Tables.getcolumn(b, col) for col in a.cols)
+  else
+    false
+  end
+end
+
+function Base.show(io::IO, ts::TableSelection)
+  println(io, "TableSelection")
+  pretty_table(io, ts, vcrop_mode=:middle)
+end
+
+# Tables.jl interface
 Tables.istable(::Type{<:TableSelection}) = true
 Tables.columnaccess(::Type{<:TableSelection}) = true
-Tables.columns(ts::TableSelection) = ts
-Tables.columnnames(ts::TableSelection) = ts.cols
-Tables.getcolumn(ts::TableSelection, col::Int) =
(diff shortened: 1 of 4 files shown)


src/transforms/select.jl
```diff
@@ -2,6 +2,38 @@
...
+end
+
+Tables.istable(::Type{<:TableSelection}) = true
+Tables.columnaccess(::Type{<:TableSelection}) = true
+Tables.columns(ts::TableSelection) = ts
+Tables.columnnames(ts::TableSelection) = ts.cols
+Tables.getcolumn(ts::TableSelection, col::Int) =
+  Tables.getcolumn(ts.table, ts.cols[col])
+Tables.getcolumn(ts::TableSelection, col::Symbol) = 
+  Tables.getcolumn(ts.table, col)
```
juliohm, line 17: Let's call the argument name `t`, the name `ts` can be associated with a collection of objects like we do with `xs` and `ys` for example.

src/transforms/select.jl
```diff
@@ -2,6 +2,38 @@
...
+function Tables.schema(ts::TableSelection)
+  schema = Tables.schema(ts.table)
+  tbtypes = schema.types
+  tbnames = collect(schema.names)
+  selecttypes = tbtypes[indexin(ts.cols, tbnames)]
+  return Tables.Schema(ts.cols, selecttypes)
+end
+
+Tables.materializer(::Type{TableSelection{T}}) where {T} = 
+  Tables.materializer(T)
```
juliohm, line 28: Can you please confirm that the materializer of the parent type is what we want here? I thought that the materializer function had to return the constructor of the type in question, in this case a TableSelection. But it may make sense to return the constructor of the parent given that the selection is a subtable always? We need to double check the API docs.

src/transforms/select.jl
```diff
@@ -2,6 +2,38 @@
...
+Tables.getcolumn(ts::TableSelection, col::Symbol) = 
+  Tables.getcolumn(ts.table, col)
+
+function Tables.schema(ts::TableSelection)
+  schema = Tables.schema(ts.table)
+  tbtypes = schema.types
+  tbnames = collect(schema.names)
+  selecttypes = tbtypes[indexin(ts.cols, tbnames)]
+  return Tables.Schema(ts.cols, selecttypes)
+end
```
juliohm, line 25:
```suggestion
function Tables.schema(t::TableSelection)
  schema = Tables.schema(t.table)
  names = schema.names
  types = schema.types
  inds = indexin(t.cols, collect(names))
  Tables.Schema(t.cols, types[inds])
end
```

src/transforms/select.jl
```diff
@@ -2,6 +2,38 @@
...
+
+Tables.materializer(::Type{TableSelection{T}}) where {T} = 
+  Tables.materializer(T)
+
+Base.:(==)(a::TableSelection, b::TableSelection) = a.table == b.table
+
```
juliohm, line 31:
```suggestion
```
We do not need this comparison operator defined, also it is not correct  given that selections are only equal if the columns selected are equal. Let's erase this definition here.

src/transforms/select.jl
```diff
@@ -62,12 +94,14 @@ function apply(transform::Select, table)
   rcols = [Tables.getcolumn(cols, name) for name in reject]
 
   # table with selected columns
-  𝒯 = (; zip(select, scols)...)
-  stable = 𝒯 |> Tables.materializer(table)
+  # 𝒯 = (; zip(select, scols)...)
+  # stable = 𝒯 |> Tables.materializer(table)
```
juliohm, line 98: Please erase commented lines in PRs.

src/transforms/select.jl
```diff
@@ -62,12 +94,14 @@ function apply(transform::Select, table)
...
 
-  stable, (reject, rcols, sperm, rinds)
+  TableSelection(table, select), (reject, rcols, sperm, rinds)
 end
 
+revert(::Select, newtable::TableSelection, cache) = newtable.table
+
```
juliohm, line 104: Do we need to add a method for TableSelection? I thought that we would leave everything else untouched because TableSelection is a valid Tables.jl table.

test/transforms.jl
```diff
@@ -12,60 +12,60 @@
 
     T = Select(:f, :d)
     n, c = apply(T, t)
-    @test Tables.columnnames(n) == (:f, :d)
+    @test Tables.columnnames(Table(n)) == (:f, :d)
```
juliohm, line 15: This shouldn't be necessary. You mean we should return a Vector instead of a Tuple? What the API docs say for Tables.columnnames? Should it return any iterable? Tuples, Vectors, etc.?

juliohm (changes requested):

src/transforms/select.jl
eliascarv: Done!

src/transforms/select.jl
eliascarv: Done!

test/transforms.jl
eliascarv: Resolved in commit af43f28.

src/transforms/select.jl
eliascarv: This method is used for a faster revert function call.

src/transforms/select.jl
eliascarv: Resolved in commit af43f28.

src/transforms/select.jl
eliascarv: Resolved in commit af43f28.

src/transforms/select.jl
```diff
@@ -2,6 +2,46 @@
...
+  cols::Vector{Symbol}
+  function TableSelection(table::T, cols::Vector{Symbol}) where {T}
+    @assert cols ⊆ Tables.columnnames(table)
+    new{T}(table, cols)
+  end
+end
+
+function Base.:(==)(a::TableSelection, b::TableSelection)
+  a.cols != b.cols && return false
+  all(Tables.getcolumn(a, col) == Tables.getcolumn(b, col) for col in a.cols)
```
juliohm, line 16: Should Tables.getcolumn be called in the output of Tables.columns instead?

src/transforms/select.jl
```diff
@@ -2,6 +2,46 @@
...
+
+function Base.show(io::IO, ts::TableSelection)
+  println(io, "TableSelection")
+  pretty_table(io, ts, vcrop_mode=:middle)
+end
+
+# Tables.jl interface
+Tables.istable(::Type{<:TableSelection}) = true
+Tables.columnaccess(::Type{<:TableSelection}) = true
+Tables.columns(t::TableSelection) = t
```
juliohm, line 27: I understand that the TableSelection already implements the AbstractColumns interface? That is why you applied Tables.getcolumn directly to TableSelection in the implementation ==?

src/transforms/select.jl
```diff
@@ -42,6 +82,8 @@ function apply(transform::Select, table)
   select  = _select(transform.cols, allcols)
   reject  = setdiff(allcols, select)
 
+  @assert select ⊆ Tables.columnnames(table)
```
juliohm, line 85: Maybe we should also assert that the select is not empty? I think this is still possible right?

src/transforms/select.jl
```diff
@@ -55,19 +97,14 @@ function apply(transform::Select, table)
...
-  # table with selected columns
-  𝒯 = (; zip(select, scols)...)
-  stable = 𝒯 |> Tables.materializer(table)
-
-  stable, (reject, rcols, sperm, rinds)
+  TableSelection(table, select), (reject, rcols, sperm, rinds)
 end
 
+revert(::Select, newtable::TableSelection, cache) = newtable.table
+
```
juliohm, line 107: Move this method to after the general revert method and add a one-line comment explaining that this is for method for performance optimization only.

test/transforms.jl
```diff
@@ -96,16 +96,24 @@
...
     tₒ = revert(T, n, c)
     @test t == tₒ
+
+    # throws
+    @test_throws AssertionError apply(Select(:x3, :y3), t)
+    @test_throws AssertionError apply(Select([:x3, :y3]), t)
+    @test_throws AssertionError apply(Select((:x3, :y3)), t)
+    @test_throws AssertionError apply(Select("x3", "y3"), t)
+    @test_throws AssertionError apply(Select(["x3", "y3"]), t)
+    @test_throws AssertionError apply(Select(("x3", "y3")), t)
```
juliohm, line 116: 👍🏽

test/transforms.jl
```diff
@@ -203,18 +211,48 @@
...
+
+    # getcolumn
+    cols = Tables.columns(t)
+    @test Tables.getcolumn(s, :a) == Tables.getcolumn(cols, :a)
+    @test Tables.getcolumn(s, 1) == Tables.getcolumn(cols, 1)
+    @test Tables.getcolumn(s, 3) == Tables.getcolumn(cols, :e)
+
+    # throws
+    @test_throws AssertionError TableTransforms.TableSelection(t, [:a, :b, :z])
+  end
```
juliohm, line 254: Amazing software engineering skills. That really helps in the long term.

juliohm (changes requested):

src/transforms/select.jl
eliascarv: Exactly!

test/transforms.jl
eliascarv: Thanks!

src/transforms/select.jl
eliascarv: Done!
Resolved in 6199b20.

src/transforms/select.jl
eliascarv: Done!
Resolved in 6199b20.

juliohm (approved):

juliohm:
Another excellent PR 👍🏽

PR #25, opened 2022-03-18
Added DropMissing transform

ceferisbarov:
I added `DropMissing` feature of `TableOperations.jl` to `TableTransforms.jl`. I added helper functions, too, so `TableOperations.jl` is not a dependency.
  
Helper functions can also be moved to a different file. There are probably too many problems with this PR. I am still trying to understand overall design of the package. Any feedback is appreciated.

P.S: Second commit simply removes a dependency that I had added by mistake.


src/transforms/dropmissing.jl
```diff
@@ -0,0 +1,82 @@
...
+    f::F
+    x::T
+end
+
+"""
+    TableOperations.filter(f, source) => TableOperations.Filter
+    source |> TableOperations.filter(f) => TableOperations.Filter
+Create a lazy wrapper that satisfies the Tables.jl interface and keeps the rows where `f(row)` is true.
+"""
+function filter end
```
juliohm, line 29: We can't simply copy/paste code from other packages over here. Please submit a separate PR for the Filter transform with tests and then after that we can look again into the DropMissing transform.

ceferisbarov: Got it, thanks. I added a filter transform. I will submit the PR as soon as I am done writing unit tests.
  
I have a slightly off-topic question: Are TableOperations and TableTransforms packages somehow coordinated? It looks like a part of their functionality is going to overlap (now that we are beginning to non-numerical tranformations like Filter and DropMissing).

juliohm:
TableTransforms.jl is a sucessor with design issues fixed.

Em sex., 18 de mar. de 2022 12:08, Jafar Isbarov ***@***.***>
escreveu:

PR #26, opened 2022-03-18
Added Filter transform

ceferisbarov:
The PR is not complete. Filter transform is not revertible (yet). I would appreciate feedback on unit tests (I will add more in the next commit).

Edit: It is revertible now. I have added more unit tests.


src/TableTransforms.jl
```diff
@@ -45,6 +45,7 @@ export
   PCA, DRS, SDS,
   Sequential,
   Parallel,
+  Filter,
```
juliohm, line 48: Please move this to after `Rename`.

src/transforms.jl
```diff
@@ -221,4 +221,5 @@ include("transforms/quantile.jl")
 include("transforms/functional.jl")
 include("transforms/eigenanalysis.jl")
 include("transforms/sequential.jl")
-include("transforms/parallel.jl")
\ No newline at end of file
+include("transforms/parallel.jl")
+include("transforms/filter.jl")
```
juliohm, line 225: Similarly, please move this to after rename.jl. The last transforms should be sequential and parallel because they are meta-transforms.

src/transforms/filter.jl
```diff
@@ -0,0 +1,51 @@
...
+    f::Base.Callable
+
+
+    function Filter(new_filter)
+        new(new_filter)
+    end
+
+    function Filter()
+        new(_default_filter)
+    end
```
juliohm, line 20: Please use the code style of the package. 2 whitespaces for indentation, no double white lines, etc.

juliohm: Also, please define outer constructors instead of inner constructors and make the field `f` of type F as a type parameter.

juliohm: Please rename `f` to `pred` for predicate function.

src/transforms/filter.jl
```diff
@@ -0,0 +1,51 @@
...
+end
+
+isrevertible(::Type{Filter}) = true
+
+#Ideally, this function would also work without a given function, as follows:
+#[item for item in iterable if item]
+function apply(transform::Filter, table)
+    indices = [i for i in range(1,length(table)) if transform.f(table[i])]
```
juliohm, line 28: `length` is not a valid function in the Tables.jl API. Your tests are failing on Linux, probably because of that.

src/transforms/filter.jl
```diff
@@ -0,0 +1,51 @@
...
+    table[indices], (indices, table[_minus(indices, table)])
```
juliohm, line 29: Similarly, you cannot index a Tables.jl table. Can you please review their API before continuing the work in the PR?

src/transforms/filter.jl
```diff
@@ -0,0 +1,51 @@
...
+end
+
+
```
juliohm, line 32: Double whitelines

src/transforms/filter.jl
```diff
@@ -0,0 +1,51 @@
...
+function revert(::Type{Filter}, newtable, cache)
+    indices = copy(cache[1])
+    org_table = copy(cache[2])
```
juliohm, line 35: Mixed naming conventions, We avoid underscores as much as possible.

src/transforms/filter.jl
```diff
@@ -0,0 +1,51 @@
...
+
+    for i in range(1, length(indices))
+        insert!(org_table, indices[i], newtable[i])
+    end
+
+    org_table
+end
+
+# Exclude the given indices
+_minus(indx, x) = setdiff(1:length(x), indx)
```
juliohm, line 45: No need to write this as a function, it is too small and easy to copy and paste.

src/transforms/filter.jl
```diff
@@ -0,0 +1,51 @@
...
+
+# Default filter to be used when no input function is given
+# Returns true if any element in the row evaluates to true
+function _default_filter(row)
+    return any(row)
+end
```
juliohm, line 51: We can probably get rid of this function as well, and use the outer constructors of `Filter` to provide the default predicate.

test/transforms.jl
```diff
@@ -654,6 +654,50 @@
...
   end
 
+  @testset "Filter" begin
+    # Test with _check_no_missing_in_row
+    col1 = [1, 2, 3, 4, 5]
+    col2 = [missing, 2, 3, 4, 5]
+    col3 = [5, 5, 5, 5, 5]
+    col4 = [1, 2, 3, 4, 5]
+    col5 = [1, 2, missing, 4, 5]
+    table = Table(;col1, col2, col3, col4, col5)
```
juliohm, line 664: Please follow the convention in the other tests. We usually give columns different names x1, x2, ..., please just adopt the convention in the nearby tests.

test/transforms.jl
```diff
@@ -654,6 +654,50 @@
...
+
+    function _check_no_missing_in_row(row)
+      for el in row
+          el === missing && return false
+      end
+  
+      return true
+    end
```
juliohm, line 672: What is this function doing here in the tests?

juliohm: I see that you are using it once, but you shouldn't call it _check_no_missing_in_row because this is a convention used for internal functions. Try to be more pedantic with code style.

juliohm (changes requested):
Attached a first review. Please make sure that tests are passing in all OSes before requesting a new review.

src/transforms/filter.jl
ceferisbarov: > Also, please define outer constructors instead of inner constructors and make the field `f` of type F as a type parameter.

Are these two for the sake of consistency or is there a practical reason?

src/transforms/filter.jl
ceferisbarov: I had written the tests with `TypedTables` and all of them passed on my machine (Ubuntu 20.04). Now I realize these two packages are not necessarily compatible.  
  
Adding `table = Tables.rowtable(input_table)` line allows us to use both the `length` function and indexing, although this always returns a `Vector{NamedTuple}` regardless of the input type. I am not sure if this is the desired outcome.

ceferisbarov: I found the reason why test run fails. I am using Julia 1.7 while the PR runs are on v1.6. I download v.16.5 and tried this, and the difference in behavior does exist. What do I do in this case? Write a code that runs with v1.6 or stick with v1.7?

src/transforms/filter.jl
```diff
@@ -0,0 +1,42 @@
...
+
+# Default filter to be used when no input function is given
+# Returns true if any element in the row evaluates to true
+Filter() = Filter(row -> any(row))
+
+isrevertible(::Type{Filter}) = true
+
+function apply(transform::Filter, input_table)
+  # Converting to Tables.rowtable to allow length function and indexing 
+  table = Tables.rowtable(input_table)
```
juliohm, line 22: Do we really need to convert to rowtable? Can't we just loop over the Tables.rows(table) iterator and check if the given row satisfies the predicate? If it does, then we can increment a counter and save the row, otherwise we just increment the counter.

src/transforms/filter.jl
```diff
@@ -0,0 +1,42 @@
...
+
+  # Get indices of the desired rows
+  indices = [i for i in range(1,length=length(table)) if transform.pred(table[i])]
```
juliohm, line 25: Try to split your lines of code into shorter lines with intermediate variables for improved readability. There is too much going on in a single line, it also makes it harder to debug bugs in the future.

test/transforms.jl
```diff
@@ -328,6 +328,50 @@
...
   end
 
+  @testset "Filter" begin
+    # Test with no_missing_in_row
+    x1 = [1, 2, 3, 4, 5]
+    x2 = [missing, 2, 3, 4, 5]
+    x3 = [5, 5, 5, 5, 5]
+    x4 = [1, 2, 3, 4, 5]
+    x5 = [1, 2, missing, 4, 5]
+    table = Tables.table([x1 x2 x3 x4 x5])
```
juliohm, line 338: Please use `t = Table(; x1, x2, x3, x4, x5)` instead like in other tests. This is a table type from TypedTables.jl.

test/transforms.jl
```diff
@@ -328,6 +328,50 @@
...
+    # Test revert function
+    newtable, cache = apply(Filter(no_missing_in_row), table)
+    reverted_table = revert(Filter, newtable, cache)
+    @test isequal(Tables.rowtable(table), reverted_table)
+
+    # Repeating the revert function
+    newtable, cache = apply(Filter(no_missing_in_row), table)
+    reverted_table_1 = revert(Filter, newtable, cache)
+    reverted_table_2 = revert(Filter, newtable, cache)
+    @test isequal(reverted_table_1, reverted_table_2)
```
juliohm, line 364: Please follow the naming convention in other tests and give these variables short single-letter names.

test/transforms.jl
```diff
@@ -328,6 +328,50 @@
...
+    x5 = [1, 2, missing, 4, 5]
+    table = Tables.table([x1 x2 x3 x4 x5])
+
+    function no_missing_in_row(row)
+      for el in row
+          el === missing && return false
+      end
+  
+      return true
+    end
```
juliohm, line 346: Call this `f` and add a comment before it explaining what it does.

juliohm (changes requested):

src/transforms/filter.jl
```diff
@@ -0,0 +1,42 @@
...
+
+The transform that filters the rows based on a given function.
+"""
+struct Filter{F} <: Transform
+  pred::F
+end
+
+# Default filter to be used when no input function is given
+# Returns true if any element in the row evaluates to true
+Filter() = Filter(row -> any(row))
```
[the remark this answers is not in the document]
ceferisbarov: I took the idea of a default filter from Python's filter function. But that function applies to list (tuples, etc.), while ours applies to tables, so I had to add the any() part. We can change it to a more a reasonable default behavior, or simply remove it.

src/transforms/filter.jl
ceferisbarov: Changing lines 24-25 to this: 
```{julia}
  l = length(table)
  f = transform.pred

  # Get indices of the desired rows
  indices = [i for i in range(1,length=l) if f(table[i])]
```

src/transforms/filter.jl
```diff
@@ -0,0 +1,44 @@
...
+
+The transform that filters the rows based on a given function.
+"""
+struct Filter{F} <: Transform
+  pred::F
+end
+
+isrevertible(::Type{Filter}) = true
+
+function apply(transform::Filter, input_table)
```
juliohm, line 16: Again, we don't use underscores in variable names. Please name the input table as `table` like in all other transforms.

src/transforms/filter.jl
```diff
@@ -0,0 +1,44 @@
...
+  # Converting to Tables.rowtable to allow length function and indexing 
+  table = Tables.rows(input_table)
+
+  l = length(table)
```
juliohm, line 20: Are you certain that the length function is part of the row table API? I thought that a row table was only iterable without a known length.

src/transforms/filter.jl
```diff
@@ -0,0 +1,44 @@
...
+  # Get indices of the desired rows
+  indices = [i for i in range(1,length=l) if f(table[i])]
+
+  # Return:
+  # - the desired rows as the new table,
+  # - indices of these rows and the remaining rows as cache
+  new_table = table[indices]
+  new_table |> Tables.materializer(input_table)
+  rem_rows = table[setdiff(1:length(table), indices)]
+  new_table, (indices, rem_rows)
```
juliohm, line 32: Again, not following the code style.

src/transforms/filter.jl
```diff
@@ -0,0 +1,44 @@
...
+
+function revert(::Type{Filter}, newtable, cache)
+  indices = copy(cache[1])
+  orgtable = copy(cache[2])
+
+  for i in range(1, length=length(indices))
+    insert!(orgtable, indices[i], newtable[i])
+  end
+
+  orgtable
```
juliohm, line 43: This code is unnecessarily complicated. We rarely need explicit calls to `copy` in Julia. Can youy please refactor and return a new table that is filled row by row with the saved indices?

juliohm (changes requested):

src/transforms/filter.jl
ceferisbarov: I have tested it and it does work, although I don't know whether it is intended to be used. Do you think I should replace it with something else?

src/transforms/filter.jl
ceferisbarov: `copy` was necessary, otherwise cache variable was changing every time I called the revert function. As a result, every revert function creates even bigger table, which obviously is not the goal. I can delete the first two lines and rewrite the function as follows:
```{julia}
function revert(::Type{Filter}, newtable, cache)
  for i in range(1, length=length(cache[1]))
    insert!(cache[2], cache[1][i], newtable[i])
  end

  cache[2]
end
```

juliohm:
Sorry @ceferisbarov , but I can't continue reviewing this PR. Please study the Tables.jl API a bit more before considering contributing to the package.

ceferisbarov:
I understand, thanks for your time.

PR #28, opened 2022-03-29
Filter and DropMissing

eliascarv:
This PR add Filter and DropMissing transformations.


src/TableTransforms.jl
```diff
@@ -30,6 +30,8 @@ export
   colapply, colrevert,
 
   # built-in
+  Filter,
+  DropMissing,
```
juliohm, line 34: Move to after Reject

src/transforms.jl
```diff
@@ -211,6 +211,7 @@ end
 # IMPLEMENTATIONS
 # ----------------
 
+include("transforms/filter.jl")
```
juliohm, line 214: Move to after select.jl

src/transforms/filter.jl
```diff
@@ -0,0 +1,60 @@
...
+end
+
+isrevertible(::Type{<:Filter}) = true
+
+rowvalues(row, colnames) = [Tables.getcolumn(row, col) for col in colnames]
+
+function apply(transform::Filter, table)
+  tablerows = Tables.rows(table)
+  colnames = Tables.columnnames(table)
+  rows = [rowvalues(row, colnames) for row in tablerows]
```
juliohm, line 21: If we need to collect the rows anyways, then we can simply use the Tables.rowtable helper function.

src/transforms/filter.jl
```diff
@@ -0,0 +1,60 @@
...
+  rinds = Vector{Int}()
+  
+  for (i, row) in enumerate(rows)
+    if transform.func(row)
+      push!(srows, row)
+    else
+      push!(rrows, row)
+      push!(rinds, i)
+    end
+  end
```
juliohm, line 35: This could be replaced by a findall or filter call with arrays of tuples.

juliohm (changes requested):

src/transforms.jl
eliascarv: OK

src/transforms/filter.jl
eliascarv: OK, I'll do it.

src/transforms/filter.jl
eliascarv: OK, I'll do it.

test/runtests.jl
```diff
@@ -27,6 +27,12 @@ end
 Polynomial(args::T...) where {T<:Real} = Polynomial(collect(args))
 (p::Polynomial)(x) = sum(a * x^(i-1) for (i, a) in enumerate(p.coeffs))
 
+function testmissing(vec1, vec2)
+  r1 = map(!ismissing, vec1) == map(!ismissing, vec2)
+  r2 = findall(ismissing, vec1) == findall(ismissing, vec2)
+  r1 && r2
```
juliohm, line 33: Can you elaborate on why we need these two checks instead of just the last with findall?

juliohm:
As usual a great PR 👍🏽 we just need to double check the testmissing helper function, maybe we could avoid introducing a function if all we need is to check the positions of the missing values with findall.

test/runtests.jl
eliascarv: This test pass without first check:
```julia
[missing, 1, 2] == [missing, 2, 1]
```
And this test pass without second check:
```julia
[missing, 1, 2] == [1, 2, missing]
```

juliohm: Thanks, I guess we could rewrite this helper function with a for loop like in the previous PR where we first check if the two entries are not missing and then compare the values?

juliohm:
Also, we need to add these two new transforms to the README table 👍🏽

test/runtests.jl
eliascarv: Sorry, I messed up. Is no `map` is `filter`.

eliascarv: Resolved with [0a3319d](https://github.com/JuliaML/TableTransforms.jl/pull/28/commits/0a3319d91a86cd4310f326976725eb82f7fadba2)

eliascarv:
This PR is ready for a final review @juliohm!

README.md
```diff
@@ -147,6 +147,8 @@ Please check the docstrings for additional information.
 |-----------|-------------|
 | `Select` | Column selection |
 | `Reject` | Column rejection |
+| `Filter` | Filter rows |
```
juliohm, line 150:
```suggestion
| `Filter` | Row filtering |
```

README.md
```diff
@@ -147,6 +147,8 @@ Please check the docstrings for additional information.
...
+| `DropMissing` | Drop rows with missings |
```
juliohm, line 151:
```suggestion
| `DropMissing` | Drop missings |
```

juliohm (commented):
I think we are ready for merging. I will adjust the descriptions in the README.

juliohm (approved):

PR #29, opened 2022-03-30
Generalize the ColSpec interface

eliascarv:
Generalize the ColSpec type to be used in the transformations: Select, Reject, DropMissing and in future transformations.


test/transforms.jl
```diff
@@ -107,8 +123,7 @@
     tₒ = revert(T, n, c)
     @test t == tₒ
 
-    # throws: Select without arguments
-    @test_throws ArgumentError Select()
```
juliohm, line 111: What is the current meaning of Select() ?

src/transforms/utils.jl
```diff
@@ -0,0 +1,15 @@
...
+# Licensed under the MIT License. See LICENSE in the project root.
+# ------------------------------------------------------------------
+
+const ColSelector = Union{Symbol, Integer, AbstractString}
+const ColSpec = Union{Vector{T}, NTuple{N, T}, Regex, Colon} where {N, T<:ColSelector}
+
+_select(colspec::Vector{Symbol}, allcols) = colspec
+_select(colspec::Vector{<:Integer}, allcols) = allcols[colspec]
+_select(colspec::Vector{<:AbstractString}, allcols) = Symbol.(colspec)
+_select(colspec::NTuple{N, <:ColSelector}, allcols) where {N} =
```
juliohm, line 11:
```suggestion
_select(colspec::NTuple{N,<:ColSelector}, allcols) where {N} =
```

src/transforms/utils.jl
```diff
@@ -0,0 +1,15 @@
+# ------------------------------------------------------------------
+# Licensed under the MIT License. See LICENSE in the project root.
+# ------------------------------------------------------------------
+
+const ColSelector = Union{Symbol, Integer, AbstractString}
+const ColSpec = Union{Vector{T}, NTuple{N, T}, Regex, Colon} where {N, T<:ColSelector}
```
juliohm, line 6:
```suggestion
const ColSpec = Union{Vector{T}, NTuple{N,T}, Regex, Colon} where {N,T<:ColSelector}
```

juliohm (commented):
It would be nice to document a bit further the utils.jl file so that first-time readers can figure out what is happening. Also, I left some comments on specific lines.

src/transforms/colspec.jl
```diff
@@ -0,0 +1,62 @@
...
+# because Vector is more flexible than Tuple.
+#
+# note 2: allcols can be of type Vector or Tuple,
+# in some cases it doesn't matter, but in others it does.
+#
+# note 3: all other methods of the function convert their arguments
+# to call the method below (main method), except the last one.
+
+# _filter(::Vector{Symbol}, allcols) - main method
+# second argument can be a Vector or Tuple
```
juliohm, line 23: I believe that this comment section can be omitted, the code already provides this information.

src/transforms/colspec.jl
```diff
@@ -0,0 +1,62 @@
...
+function _filter(colspec::Vector{Symbol}, allcols)
```
juliohm, line 24: Please rename allcols to cols in all places

juliohm (changes requested):
I believe that this PR is already mixing up too many things. If you can split it up in a PR with the minimum amount of changes to introduce a ColSpec type, we can then write separate PRs updating the Select/Reject, Rename, DropMissing, ...

eliascarv:
I think I'll close this PR and make a new one that just adds the colspec.jl file without changing the Select, Revert and DropMissing.
What do you think?

eliascarv:
This PR will be closed and replaced with smaller PRs.

PR #30, opened 2022-03-31
ColSpec

eliascarv:
This PR makes the ColSpec interface more generic so it can be used in all transformations that use column selections.

### DIFF
--- a/src/transforms/colspec.jl
+++ b/src/transforms/colspec.jl
@@ -5,7 +5,7 @@
 # types used to select a column
 const ColSelector = Union{Symbol, Integer, AbstractString}
 # types used to filter columns
-const ColSpec = Union{Vector{T}, NTuple{N,T}, Regex, Colon} where {N,T<:ColSelector}
+# const ColSpec = Union{Vector{T}, NTuple{N,T}, Regex, Colon} where {N,T<:ColSelector}
 
 # filter table columns using colspec
 function _filter(colspec::Vector{Symbol}, cols)
--- a/src/transforms/colspec.jl
+++ b/src/transforms/colspec.jl
@@ -3,9 +3,7 @@
 # ------------------------------------------------------------------
 
 # types used to select a column
-const ColSelector = Union{Symbol, Integer, AbstractString}
-# types used to filter columns
-# const ColSpec = Union{Vector{T}, NTuple{N,T}, Regex, Colon} where {N,T<:ColSelector}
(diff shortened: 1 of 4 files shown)


src/transforms/colspec.jl
```diff
@@ -0,0 +1,39 @@
+# ------------------------------------------------------------------
+# Licensed under the MIT License. See LICENSE in the project root.
+# ------------------------------------------------------------------
+
+# types used to select a column
+const ColSelector = Union{Symbol, Integer, AbstractString}
```
juliohm, line 6:
```suggestion
const ColSelector = Union{Symbol,Integer,AbstractString}
```

test/transforms.jl
```diff
@@ -1,6 +1,92 @@
 @testset "Transforms" begin
   # using MersenneTwister for compatibility between Julia versions
   rng = MersenneTwister(42)
+  @testset "colspec" begin
```
juliohm, line 4:
```suggestion
  @testset "ColSpec" begin
```

src/transforms/colspec.jl
```diff
@@ -0,0 +1,39 @@
...
+# types used to filter columns
+# const ColSpec = Union{Vector{T}, NTuple{N,T}, Regex, Colon} where {N,T<:ColSelector}
```
juliohm, line 8:
```suggestion
```

juliohm (changes requested):
Amazing PR as usual. Just a few adjustments before we can merge it.

src/transforms/colspec.jl
eliascarv: Resolved with [b34218e](https://github.com/JuliaML/TableTransforms.jl/pull/30/commits/b34218ebceae1e22b71d02dc6c27176fc490f1a3)

test/transforms.jl
eliascarv: Resolved with [b34218e](https://github.com/JuliaML/TableTransforms.jl/pull/30/commits/b34218ebceae1e22b71d02dc6c27176fc490f1a3)

src/transforms/colspec.jl
eliascarv: Resolved with [b34218e](https://github.com/JuliaML/TableTransforms.jl/pull/30/commits/b34218ebceae1e22b71d02dc6c27176fc490f1a3)

eliascarv:
This PR is done to final review.

PR #33, opened 2022-04-03
Implementing ColSpec interface in DropMissing

eliascarv:
This PR implements the ColSpec interface in DropMissing. This is the last part of the ColSpec interface implementation in the current code.

### DIFF
--- a/src/transforms/filter.jl
+++ b/src/transforms/filter.jl
@@ -60,7 +60,6 @@ struct DropMissing{S<:ColSpec} <: Stateless
   colspec::S
 end
 
-# argument error
 DropMissing(::Tuple{}) = throw(ArgumentError("Cannot create a DropMissing object with empty tuple."))
 
 DropMissing() = DropMissing(:)
--- a/src/transforms/filter.jl
+++ b/src/transforms/filter.jl
@@ -42,6 +42,7 @@ const VecOrTuple{T} = Union{Vector{T}, NTuple{N, T}} where {T, N}
 
 """
     DropMissing()
+    DropMissing(:)
 
 Drop all rows with missing values in table.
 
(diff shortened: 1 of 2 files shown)


src/transforms/filter.jl
```diff
@@ -55,20 +56,36 @@ Drop all rows with missing values in selects columns `col₁`, `col₂`, ..., `c
...
 Drop all rows with missing values in columns that match with `regex`.
 """
-DropMissing() = Filter(row -> all(!ismissing, row))
+struct DropMissing{S<:ColSpec} <: Stateless
+  colspec::S
+end
 
-DropMissing(cols::VecOrTuple{T}) where {T<:Union{Symbol, Integer}} =
-  Filter(row -> all(!ismissing, getindex.(Ref(row), cols)))
+# argument error
```
juliohm, line 63:
```suggestion
```

src/transforms/filter.jl
```diff
@@ -55,20 +56,36 @@ Drop all rows with missing values in selects columns `col₁`, `col₂`, ..., `c
...
 
-function DropMissing(regex::Regex)
-  Filter() do row
-    cols = _filter(regex, propertynames(row))
-    all(!ismissing, getindex.(Ref(row), cols))
-  end
+isrevertible(::Type{<:DropMissing}) = true
+
+_ftrans(::DropMissing{Colon}, table) =
+  Filter(row -> all(!ismissing, row))
```
juliohm, line 74: Can we avoid introducing this helper function, and instead use the DropMissing constructor to return a Filter object?

juliohm (changes requested):

src/transforms/filter.jl
eliascarv: I don't think it makes sense to use the constructor of one type to return an object of another type.
But we can use Functors.

juliohm: Agree, the functor interface is already provided as a fallback to all Transform types. The helper function seems like the best alternative.

juliohm (approved):

Issue #27, opened 2022-03-18
Dealing with missing values


ceferisbarov:
I am not sure whether this is intentional, but when the input table has missing values, `assert_continuous` function throws `AssertionError: columns must hold continuous variables`, even though all other values are continuous.
  
Considering that there is still no transform for imputation, I think it would make sense to add an option to skip the missing values.

ablaom:
```julia
julia> scitype([1.0, 2.3, missing, 3.5])
AbstractVector{Union{Missing, Continuous}} (alias for AbstractArray{Union{Missing, Continuous}, 1})
```

ceferisbarov:
I generalized assert_continuous to work in the presence of missing values, but this results in errors during the transformation process. For example:
```
MethodError: Cannot `convert` an object of type Missing to an object of type Float64
```
I can think of several alternatives:
- Automatically deal with missing values before the transformation to avoid these errors
- Simply generalize assert_continuous and allow rest of the transform to throw these errors
- Create a helpful warning message in assert_continuous function
- Do nothing and let assert_continuous to assert that the table is not continuous.
  
Should I implement one of these? Or maybe something else?

juliohm:
That is a great suggestion @ablaom , if I understood correctly you are suggesting that we use scitype on the column and then check if the result is <: AbstractVector{Union{Missing,Continuous}} ?

juliohm:
@ceferisbarov try to use scitype as @ablaom suggested, happy to review a PR.

ceferisbarov:
@juliohm Great, I am working on it. The problem is that  `AbstractVector{Continuous} <: AbstractVector{Union{Missing,Continuous}}` returns false, but
 `Continuous <: Union{Missing, Continuous}` returns true. Is there a reason why I shouldn't use the latter? The following line:
```{julia}
@assert all(T <: Continuous for T in types) "columns must hold continuous variables"
```
would be replaced with:
```{julia}
@assert all(T <: Union{Missing, Continuous} for T in types) "columns must hold continuous variables"
```

juliohm:
I think we need two distinct assertion functions. The one we have is more strict in the sense that it asserts that we don't have missing values. Maybe we should add a new function assert_continuous_or_missing if a transform supports missing values. Can you please remind me why we started discussing this generalization? Do we really need to allow missing values in our currently implemented transforms? As far as I remember none of our statistical transforms, which require continuous values, support missing values.

ceferisbarov:
[...]

PR #34, opened 2022-04-04
Support skipping missing values for certain transforms

ceferisbarov:
- Add `assert_continuous_or_missing` function which replaces `assert_continuous` in transforms that are going to support skipping missing values.
- Revise those transforms to support skipping missing values. Makes this an argument that defaults to true.
  
Opened this PR to give better idea about what I have in mind. Still working on tests.

### DIFF
--- a/src/assertions.jl
+++ b/src/assertions.jl
@@ -9,6 +9,9 @@ function assert_continuous(table)
 end
 
 function assert_continuous_or_missing(table)
-  types = schema(table).scitypes
-  @assert all(T <: Union{Missing, Continuous} for T in types) "columns must hold continuous variables"
+  cols = Tables.columns(table)
+  for col in cols
+      type = scitype(col)
+      @assert type <: AbstractVector{<:Union{Missing, Continuous}} "columns must hold continuous variables"
+  end
 end
--- a/src/assertions.jl
+++ b/src/assertions.jl
@@ -7,3 +7,8 @@ function assert_continuous(table)
   types = schema(table).scitypes
   @assert all(T <: Continuous for T in types) "columns must hold continuous variables"
 end
+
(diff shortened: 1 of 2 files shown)


src/assertions.jl
```diff
@@ -7,3 +7,11 @@ function assert_continuous(table)
   types = schema(table).scitypes
   @assert all(T <: Continuous for T in types) "columns must hold continuous variables"
 end
+
+function assert_continuous_or_missing(table)
+  cols = Tables.columns(table)
+  for col in cols
+      type = scitype(col)
+      @assert type <: AbstractVector{<:Union{Missing, Continuous}} "columns must hold continuous variables"
```
juliohm, line 15: Can you elaborate on why we are not using the same pattern of the assert_continuous function with T <: Union{<:Continuous,Missing} ?

juliohm (changes requested):
This PR is doing two things at the same time, and I am still unsure about how we should handle missing values more generally. Missing values in Julia are still "work in progress" in the sense that people feel differently about the default model imposed by the language. Different ecosystems adopted different missing value models, and maybe we should delay this discussion to after the missing values model settles in the language and its data science community.

To give you some context, we are planning a v1.0 release this month, and it won't include support for missing values. Missing values could be considered in a future v2.0 release (after a year or so playing around with different models).

juliohm:
Although we appreciate the effort, I think the best option we have right now is to close the PR and reconsider missing values in a year or so from now.

src/assertions.jl
ceferisbarov: I simply implemented the suggestion made in #27 .  I also agree that `T <: Union{<:Continuous,Missing}` is a more elegant solution.

PR #35, opened 2022-04-04
Avoid `Union{Missing,T}` in columns after DropMissing

eliascarv:
This PR adds a post-processing in the apply function of the DropMissing transformation and a pre-processing in the revert function.
This is necessary so that the type of columns where the DropMissing transformation will be applied is `T` instead of `Union{Missing,T}` .

### DIFF
--- a/src/transforms/filter.jl
+++ b/src/transforms/filter.jl
@@ -65,7 +65,6 @@ DropMissing(cols::T...) where {T<:ColSelector} =
 
 isrevertible(::Type{<:DropMissing}) = true
 
-# ftrans
 _ftrans(::DropMissing{Colon}, cols) =
   Filter(row -> all(!ismissing, row))
 
@@ -78,29 +77,35 @@ _nonmissing(::Type{Union{Missing,T}}, x) where {T} = collect(T, x)
 _nonmissing(x) = _nonmissing(eltype(x), x)
 
 function apply(transform::DropMissing, table)
-  colnames = Tables.columnnames(table)
-  selnames = _filter(transform.colspec, colnames)
+  names = Tables.columnnames(table)
+  coltypes = Tables.schema(table).types
+  selnames = _filter(transform.colspec, names)
   ftrans = _ftrans(transform, selnames)
   newtable, fcache = apply(ftrans, table)
 
   # post-processing
-  coltable = Tables.columntable(newtable)
-  pcolumns = [nm ∈ selnames ? _nonmissing(x) : x for (nm, x) in pairs(coltable)]
-  𝒯 = (; zip(colnames, pcolumns)...)
+  cols = Tables.columns(newtable)
+  pcols = map(names) do n
+    x = Tables.getcolumn(cols, n)
+    n ∈ selnames ? _nonmissing(x) : x
+  end
+  𝒯 = (; zip(names, pcols)...)
   ptable = 𝒯 |> Tables.materializer(newtable)
 
-  types = Tables.schema(table).types
-  ptable, (ftrans, fcache, types)
+  ptable, (ftrans, fcache, coltypes)
 end
 
 function revert(::DropMissing, newtable, cache)
-  ftrans, fcache, types = cache
+  ftrans, fcache, coltypes = cache
 
   # pre-processing
-  colnames = Tables.columnnames(newtable)
-  coltable = Tables.columntable(newtable)
-  pcolumns = [collect(T, x) for (T, x) in zip(types, coltable)]
-  𝒯 = (; zip(colnames, pcolumns)...)
+  cols = Tables.columns(newtable)
+  names = Tables.columnnames(newtable)
+  pcols = map(zip(coltypes, names)) do (T, n)
+    x = Tables.getcolumn(cols, n)
+    collect(T, x)
+  end
+  𝒯 = (; zip(names, pcols)...)
   ptable = 𝒯 |> Tables.materializer(newtable)
 
   revert(ftrans, ptable, fcache)
--- a/src/transforms/filter.jl
+++ b/src/transforms/filter.jl
@@ -82,38 +82,33 @@ function _nonmissing(columns, col)
   _nonmissing(eltype(c), c)
 end
 
-function _nonmissing(table, cols, allcols)
-  columns = Tables.columns(table)
+function apply(transform::DropMissing, table)
+  allcols = Tables.columnnames(table)
+  cols    = _filter(transform.colspec, allcols)
+  ftrans  = _ftrans(transform, cols)
+  newtable, fcache = apply(ftrans, table)
+
+  # post-processing
+  columns = Tables.columns(newtable)
   newcols = [col ∈ cols ? _nonmissing(columns, col) : Tables.getcolumn(columns, col)
              for col in allcols]
   𝒯 = (; zip(allcols, newcols)...)
-  𝒯 |> Tables.materializer(table)
-end
+  ptable = 𝒯 |> Tables.materializer(newtable)
 
-# reverttypes
-function _reverttypes(table, types)
-  columns = Tables.columns(table)
-  allcols = Tables.columnnames(table)
-  newcols = [collect(T, Tables.getcolumn(columns, col)) 
-             for (T, col) in zip(types, allcols)]
-  𝒯 = (; zip(allcols, newcols)...)
-  𝒯 |> Tables.materializer(table)
-end
-
-function apply(transform::DropMissing, table)
-  allcols = Tables.columnnames(table)
-  cols = _filter(transform.colspec, allcols)
-  ftrans = _ftrans(transform, cols)
-  newtable, fcache = apply(ftrans, table)
-  # post-processing
   types = Tables.schema(table).types
-  ptable = _nonmissing(newtable, cols, allcols)
   ptable, (ftrans, fcache, types)
 end
 
 function revert(::DropMissing, newtable, cache)
   ftrans, fcache, types = cache
+
   # pre-processing
-  ptable = _reverttypes(newtable, types)
+  columns = Tables.columns(newtable)
+  allcols = Tables.columnnames(newtable)
+  newcols = [collect(T, Tables.getcolumn(columns, col))
+             for (T, col) in zip(types, allcols)]
+  𝒯 = (; zip(allcols, newcols)...)
+  ptable = 𝒯 |> Tables.materializer(newtable)
+
   revert(ftrans, ptable, fcache)
 end
--- a/src/transforms/filter.jl
+++ b/src/transforms/filter.jl
@@ -27,7 +27,7 @@ function apply(transform::Filter, table)
 end
 
 function revert(::Filter, newtable, cache)
-  rows = Vector{NamedTuple}(Tables.rowtable(newtable))
+  rows = Tables.rowtable(newtable)
 
   for (i, row) in cache
     insert!(rows, i, row)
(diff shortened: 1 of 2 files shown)


juliohm:
I think this current implementation has the issue that it forces a Union{Missing,T} in the revert even when the original table doesn't have missing values. As we discussed over Zulip, we just need to save the original column eltype during the apply and reuse it in the revert.

juliohm (changes requested):
Need to fix the revert so that it restores the original eltype, and also add tests.

eliascarv:
Done. I made the requested changes and added tests @juliohm.

src/transforms/filter.jl
```diff
@@ -69,22 +65,55 @@ DropMissing(cols::T...) where {T<:ColSelector} =
...
+_nonmissing(::Type{T}, c) where {T} = c
+_nonmissing(::Type{Union{Missing,T}}, c) where {T} =
+  collect(T, c)
+
+function _nonmissing(columns, col)
+  c = Tables.getcolumn(columns, col)
+  _nonmissing(eltype(c), c)
+end
+
+function _nonmissing(table, cols, allcols)
```
juliohm, line 85: I think this is an overcomplicated design. Can we simply pass the table as the first argument and retrieve the column names inside the function instead of passing allcols? Also, try to avoid creating external functions unless that is strictly necessary. I thought that we should be able to write a compact apply and revert doing pre and post processing without all this additional helper functions.

juliohm (changes requested):

src/transforms/filter.jl
eliascarv: OK

src/transforms/filter.jl
```diff
@@ -69,22 +65,50 @@ DropMissing(cols::T...) where {T<:ColSelector} =
...
+  allcols = Tables.columnnames(table)
+  cols    = _filter(transform.colspec, allcols)
+  ftrans  = _ftrans(transform, cols)
   newtable, fcache = apply(ftrans, table)
-  newtable, (ftrans, fcache)
+
+  # post-processing
+  columns = Tables.columns(newtable)
+  newcols = [col ∈ cols ? _nonmissing(columns, col) : Tables.getcolumn(columns, col)
+             for col in allcols]
```
juliohm, line 94: Please use the do syntax to shorten up this line. The variable names in this function are a bit confusing as well. Too many cols, allcols, columns with different meanings. Can we consider other names?

src/transforms/filter.jl
```diff
@@ -69,22 +65,50 @@ DropMissing(cols::T...) where {T<:ColSelector} =
...
 function revert(::DropMissing, newtable, cache)
-  ftrans, fcache = cache
-  revert(ftrans, newtable, fcache)
+  ftrans, fcache, types = cache
+
+  # pre-processing
+  columns = Tables.columns(newtable)
+  allcols = Tables.columnnames(newtable)
+  newcols = [collect(T, Tables.getcolumn(columns, col))
+             for (T, col) in zip(types, allcols)]
```
juliohm, line 109: Try to place this into a single line or introduce new intermediate variables so that the line is short enough.

juliohm (changes requested):

src/transforms/filter.jl
eliascarv: OK

src/transforms/filter.jl
eliascarv: I used the Tables.columntable function to simplify the code.

src/transforms/filter.jl
eliascarv: I used the Tables.columntable function to simplify the code.

src/transforms/filter.jl
```diff
@@ -69,22 +65,45 @@ DropMissing(cols::T...) where {T<:ColSelector} =
...
-  allcols = Tables.columnnames(table)
-  cols = _filter(transform.colspec, allcols)
+_ftrans(::DropMissing, cols) =
   Filter(row -> all(!ismissing, getindex.(Ref(row), cols)))
-end
+
+# nonmissing 
+_nonmissing(::Type{T}, c) where {T} = c
+_nonmissing(::Type{Union{Missing,T}}, c) where {T} =
+  collect(T, c)
```
juliohm, line 78:
```suggestion
_nonmissing(::Type{Union{Missing,T}}, c) where {T} = collect(T, c)
```

src/transforms/filter.jl
```diff
@@ -69,22 +65,45 @@ DropMissing(cols::T...) where {T<:ColSelector} =
...
+
+_nonmissing(col) = _nonmissing(eltype(col), col)
 
 function apply(transform::DropMissing, table)
-  ftrans = _ftrans(transform, table)
+  colnames = Tables.columnnames(table)
+  select = _filter(transform.colspec, colnames)
```
juliohm, line 84: Maybe rename this to selnames to  indicate that these are still names of columns.

src/transforms/filter.jl
```diff
@@ -69,22 +65,45 @@ DropMissing(cols::T...) where {T<:ColSelector} =
...
+  ftrans = _ftrans(transform, select)
   newtable, fcache = apply(ftrans, table)
-  newtable, (ftrans, fcache)
+
+  # post-processing
+  coltable = Tables.columntable(newtable)
+  pcolumns = [nm ∈ select ? _nonmissing(col) : col for (nm, col) in pairs(coltable)]
```
juliohm, line 90: In other transforms we call the vector of values of a column `x`. Can you also use this letter here? Same for the implementations of _nonmissing, they should get a vector x instead of a vector c

juliohm (changes requested):

src/transforms/filter.jl
eliascarv: OK

src/transforms/filter.jl
eliascarv: OK

juliohm: Please use the like button if you want to express agreement. Don`t need to reply OK in every suggestion because these are sent to everyone's mail box.

src/transforms/filter.jl
eliascarv: Done with 7db5e28

src/transforms/filter.jl
eliascarv: Done with [7db5e28](https://github.com/JuliaML/TableTransforms.jl/pull/35/commits/7db5e28ea9f808a8a4a1f1426f7ccf7b866f034c)

src/transforms/filter.jl
eliascarv: Donde with [7db5e28](https://github.com/JuliaML/TableTransforms.jl/pull/35/commits/7db5e28ea9f808a8a4a1f1426f7ccf7b866f034c)

src/transforms/filter.jl
```diff
@@ -69,22 +65,43 @@ DropMissing(cols::T...) where {T<:ColSelector} =
 
 isrevertible(::Type{<:DropMissing}) = true
 
-_ftrans(::DropMissing{Colon}, table) =
+# ftrans
```
juliohm, line 68:
```suggestion
```

src/transforms/filter.jl
```diff
@@ -69,22 +65,43 @@ DropMissing(cols::T...) where {T<:ColSelector} =
...
-  ftrans = _ftrans(transform, table)
+  colnames = Tables.columnnames(table)
+  selnames = _filter(transform.colspec, colnames)
+  ftrans = _ftrans(transform, selnames)
   newtable, fcache = apply(ftrans, table)
-  newtable, (ftrans, fcache)
+
+  # post-processing
+  coltable = Tables.columntable(newtable)
+  pcolumns = [nm ∈ selnames ? _nonmissing(x) : x for (nm, x) in pairs(coltable)]
```
juliohm, line 88:
```suggestion
  pcolumns = [n ∈ selnames ? _nonmissing(x) : x for (n, x) in pairs(coltable)]
```

src/transforms/filter.jl
```diff
@@ -69,22 +65,43 @@ DropMissing(cols::T...) where {T<:ColSelector} =
...
 function apply(transform::DropMissing, table)
-  ftrans = _ftrans(transform, table)
+  colnames = Tables.columnnames(table)
+  selnames = _filter(transform.colspec, colnames)
+  ftrans = _ftrans(transform, selnames)
   newtable, fcache = apply(ftrans, table)
-  newtable, (ftrans, fcache)
+
+  # post-processing
+  coltable = Tables.columntable(newtable)
```
juliohm, line 87: Can't we just use Tables.columns here? Why do we need to materialize a new table type just to convert the column types?

src/transforms/filter.jl
```diff
@@ -69,22 +65,43 @@ DropMissing(cols::T...) where {T<:ColSelector} =
...
+  pcolumns = [nm ∈ selnames ? _nonmissing(x) : x for (nm, x) in pairs(coltable)]
+  𝒯 = (; zip(colnames, pcolumns)...)
+  ptable = 𝒯 |> Tables.materializer(newtable)
+
+  types = Tables.schema(table).types
```
juliohm, line 92: This should be renamed to coltypes and moved to after colnames in the beginning of the function.

src/transforms/filter.jl
```diff
@@ -69,22 +65,43 @@ DropMissing(cols::T...) where {T<:ColSelector} =
...
 end
 
 function revert(::DropMissing, newtable, cache)
-  ftrans, fcache = cache
-  revert(ftrans, newtable, fcache)
+  ftrans, fcache, types = cache
+
+  # pre-processing
+  colnames = Tables.columnnames(newtable)
+  coltable = Tables.columntable(newtable)
```
juliohm, line 101: Can we replace columntable by columns? I am finding it strange that we need to use a specific table representation just to change the column type.

test/transforms.jl
```diff
@@ -598,6 +598,45 @@
     @test isequalmissing(n.e, [missing, 5, 6, 5])
     @test isequalmissing(n.f, [4, missing, 4, 5])
 
+    # column eltype
+    ttypes = Tables.schema(t).types
```
juliohm, line 602: move this to after the apply call and remove the redundant comment

test/transforms.jl
```diff
@@ -598,6 +598,45 @@
...
+    T = DropMissing()
+    n, c = apply(T, t)
+    ntypes = Tables.schema(n).types
+    @test ntypes[1] == Int
+    @test ntypes[2] == Int
+    @test ntypes[3] == Int
+    @test ntypes[4] == Int
+    @test ntypes[5] == Int
+    @test ntypes[6] == Int
+    tₒ = revert(T, n, c)
```
juliohm, line 613: move this revert call to after the apply call

test/transforms.jl
```diff
@@ -598,6 +598,45 @@
...
+    n, c = apply(T, t)
+    ntypes = Tables.schema(n).types
+    @test ntypes[1] == Union{Missing,Int}
+    @test ntypes[2] == Int
+    @test ntypes[3] == Union{Missing,Int}
+    @test ntypes[4] == Union{Missing,Int}
+    @test ntypes[5] == Int
+    @test ntypes[6] == Int
+    tₒ = revert(T, n, c)
+    @test ttypes == Tables.schema(tₒ).types
```
juliohm, line 638: similar comments apply to this second test set

juliohm (changes requested):

src/transforms/filter.jl
eliascarv: Made in [1514725](https://github.com/JuliaML/TableTransforms.jl/pull/35/commits/151472520bcb0c3d5222738bfa33dce003172cf7)

src/transforms/filter.jl
eliascarv: Made in [1514725](https://github.com/JuliaML/TableTransforms.jl/pull/35/commits/151472520bcb0c3d5222738bfa33dce003172cf7)

src/transforms/filter.jl
eliascarv: Made in [1514725](https://github.com/JuliaML/TableTransforms.jl/pull/35/commits/151472520bcb0c3d5222738bfa33dce003172cf7)

src/transforms/filter.jl
eliascarv: Made in [1514725](https://github.com/JuliaML/TableTransforms.jl/pull/35/commits/151472520bcb0c3d5222738bfa33dce003172cf7)

test/transforms.jl
eliascarv: Made in [1514725](https://github.com/JuliaML/TableTransforms.jl/pull/35/commits/151472520bcb0c3d5222738bfa33dce003172cf7)

test/transforms.jl
eliascarv: Made in [1514725](https://github.com/JuliaML/TableTransforms.jl/pull/35/commits/151472520bcb0c3d5222738bfa33dce003172cf7)

test/transforms.jl
eliascarv: Made in [1514725](https://github.com/JuliaML/TableTransforms.jl/pull/35/commits/151472520bcb0c3d5222738bfa33dce003172cf7)

src/transforms/filter.jl
```diff
@@ -69,22 +65,48 @@ DropMissing(cols::T...) where {T<:ColSelector} =
...
-end
+
+# nonmissing 
+_nonmissing(::Type{T}, x) where {T} = x
+_nonmissing(::Type{Union{Missing,T}}, x) where {T} = collect(T, x)
+_nonmissing(x) = _nonmissing(eltype(x), x)
 
 function apply(transform::DropMissing, table)
-  ftrans = _ftrans(transform, table)
+  names = Tables.columnnames(table)
```
juliohm, line 80: given the other names in this function, maybe we should also name this colnames

juliohm: or maybe rename coltypes --> types and selnames --> selection

juliohm (changes requested):
Just a few minor changes now and I think we are ready

src/transforms/filter.jl
eliascarv: OK

eliascarv: Made in [3634072](https://github.com/JuliaML/TableTransforms.jl/pull/35/commits/36340728613133dd12ca9c6c27a04e86f9d09c1c)

eliascarv:
This PR is ready for a final review @juliohm!

juliohm (approved):

Issue #36, opened 2022-04-04
Please move colspec.jl to the root folder


juliohm:
I noticed that you added colspec.jl to the transforms folder, but it is not a transform @eliascarv. Can you please move it to the root folder and update the include?

eliascarv:
OK, I will do it

juliohm:
@eliascarv appreciate if you can take a look at this now that we merged the PR you were working on. 💯

eliascarv:
OK, I will do it now!

juliohm:
This issue is now fixed. I think we also need to rename _filter to a public function name to be used by other transforms. I will open a separate issue for that.

PR #38, opened 2022-04-09
Move colspec and rename arguments

eliascarv:
This PR moves the colspec file to src and renames `cols` to `names` for consistency with the rest of the code.


src/transforms.jl
```diff
@@ -211,7 +211,7 @@ end
 # IMPLEMENTATIONS
 # ----------------
 
-include("transforms/colspec.jl")
+include("colspec.jl")
```
juliohm, line 214: This include should go in the main TableTransforms.jl file. This list of includes here is for implementations of the Transform trait.

juliohm (changes requested):

src/transforms.jl
eliascarv: ok

eliascarv: Made in [933754d](https://github.com/JuliaML/TableTransforms.jl/pull/38/commits/933754de0e18e04bafc02872229966abf01cbfa8)

juliohm (approved):

juliohm:
Fix #36

PR #42, opened 2022-04-12
Add Replace

eliascarv:
This PR implements Replace tranformation. The `Replace(oldᵢ => newᵢ)` transformations replaces all occurrences of `oldᵢ` with `newᵢ` in the table.

Note: If the type of `old` is different from the type of `new`, the column type is changed to a type common to the new column elements, without promotion.
Example:
```julia-repl
julia> using TableTransforms, DataFrames

julia> df = DataFrame(a = [2, 1, 2], b = [1, 1, 1])
3×2 DataFrame       
 Row │ a      b     
     │ Int64  Int64 
─────┼──────────────
   1 │     2      1 
   2 │     1      1 
   3 │     2      1 

julia> T = Replace(1 => 1.5)
Replace{Int64, Float64}(IdDict(1 => 1.5))

julia> T(df)
3×2 DataFrame
 Row │ a     b       
     │ Real  Float64 
─────┼───────────────
   1 │  2        1.5
   2 │  1.5      1.5
   3 │  2        1.5
```


src/TableTransforms.jl
```diff
@@ -39,6 +39,7 @@ export
   Coalesce,
   Identity,
   Center,
+  Replace,
```
juliohm, line 42: Move it to before Identity. Everything below Identity is more "numerical".

src/transforms/replace.jl
```diff
@@ -0,0 +1,29 @@
+# ------------------------------------------------------------------
+# Licensed under the MIT License. See LICENSE in the project root.
+# ------------------------------------------------------------------
+"""
+    Replace(old₁ => new₁, old₂ => new₂, ..., oldₙ => newₙ)
+
+Replaces all occurrences of `oldᵢ` with `newᵢ` in the table.
+"""
+struct Replace{K,V} <: Colwise
+  oldnew::IdDict{K,V}
```
juliohm, line 10: Should we call this pairs for consistency with other transforms? Also, first time I see IdDict, interesting finding.

juliohm (changes requested):

src/transforms/replace.jl
eliascarv: OK.
IdDict is used for this reason:
```julia-repl
julia> Dict((1 => 3, true => false))
Dict{Integer, Integer} with 1 entry:
  true => false

julia> IdDict((1 => 3, true => false))
IdDict{Integer, Integer} with 2 entries:
  1    => 3
  true => false
```

src/TableTransforms.jl
eliascarv: OK

eliascarv: Made in [0159ff2](https://github.com/JuliaML/TableTransforms.jl/pull/42/commits/0159ff21165a22578f40d5d28e94e37a44c31b66)

juliohm (commented):
Need to resolve newly added conflicts with master.

src/TableTransforms.jl
```diff
@@ -37,6 +37,7 @@ export
   DropMissing,
   Rename,
   Coalesce,
+  Replace,
```
juliohm, line 40: Move to before Coalesce

src/transforms.jl
```diff
@@ -215,6 +215,7 @@ include("transforms/select.jl")
 include("transforms/filter.jl")
 include("transforms/rename.jl")
 include("transforms/coalesce.jl")
+include("transforms/replace.jl")
```
juliohm, line 218: Same here.

src/transforms/replace.jl
```diff
@@ -0,0 +1,30 @@
+# ------------------------------------------------------------------
+# Licensed under the MIT License. See LICENSE in the project root.
+# ------------------------------------------------------------------
+
+"""
+    Replace(old₁ => new₁, old₂ => new₂, ..., oldₙ => newₙ)
+
+Replaces all occurrences of `oldᵢ` with `newᵢ` in the table.
```
juliohm, line 8:
```suggestion
Replaces `oldᵢ` value with `newᵢ` value in the table.
```

test/transforms.jl
```diff
@@ -673,6 +673,89 @@
...
+    tₒ = revert(T, n, c)
+    @test t == tₒ
+
+    # table schema after apply and revert
+    T = Replace(1 => 1.5, 5 => 5.5, 4 => true)
+    n, c = apply(T, t)
+    tₒ = revert(T, n, c)
+    ttypes = Tables.schema(t).types
+    ntypes = Tables.schema(n).types
+    @test ntypes[1] == typejoin(typeof.(n.a)...)
```
juliohm, line 724: Isn't this the eltype of n.a? Can we just use eltype?

juliohm (changes requested):

test/transforms.jl
eliascarv: `n.a` has values with different types.
In this case:
```julia
a = Real[3, 2, 1.5, true, 5.5, 3]
typejoin(typeof.(a)...)
typejoin(Int, Int, Float64, Bool, Float64, Int)
Real
```
It's just to make sure the common type between column values ​​is correct.

src/TableTransforms.jl
eliascarv: OK

src/transforms.jl
eliascarv: OK

src/transforms/replace.jl
eliascarv: OK

eliascarv: Made in [6e3d31e](https://github.com/JuliaML/TableTransforms.jl/pull/42/commits/6e3d31efa6523dd7b8c4fb85c21d07ac9cc4e738)

src/transforms.jl
eliascarv: Made in [9c5c293](https://github.com/JuliaML/TableTransforms.jl/pull/42/commits/9c5c293c10346a2cb1eddf7f2bf11923779906de)

src/TableTransforms.jl
eliascarv: Made in [9c5c293](https://github.com/JuliaML/TableTransforms.jl/pull/42/commits/9c5c293c10346a2cb1eddf7f2bf11923779906de)

test/transforms.jl
juliohm: Yes, but eltype(n.a) returns Real as well. What am I missing?

eliascarv: `eltype(n.a)` could be `Number` or `Any`:
```julia
a = Any[3, 2, 1.5, true, 5.5, 3]
typejoin(typeof.(a)...)
typejoin(Int, Int, Float64, Bool, Float64, Int)
Real
```
This check is just to make sure the algorithm is working correctly.

src/transforms/replace.jl
```diff
@@ -0,0 +1,30 @@
...
+isrevertible(::Type{<:Replace}) = true
+
+function colcache(transform::Replace, x)
+  olds = keys(transform.pairs)
+  inds = [findall(v -> v === old, x) .=> old for old in olds]
+  Dict(reduce(vcat, inds))
+end
+
+colapply(transform::Replace, x, c) =
+  map(v -> get(transform.pairs, v, v), x)
```
juliohm, line 27: Can we use a simple list comprehension here?

src/transforms/replace.jl
```diff
@@ -0,0 +1,30 @@
...
+
+colrevert(::Replace, x, c) =
+  map(i -> get(c, i, x[i]), 1:length(x))
```
juliohm, line 30: Same here, a list comprehension can improve the readability of this line.

test/transforms.jl
juliohm: I don't understand, why we cannot use eltype in this test? What do we gain with typejoin(typeof.(...))?

juliohm (changes requested):

test/transforms.jl
eliascarv: OK, i will change the tests.

eliascarv: Made in [2f87e1d](https://github.com/JuliaML/TableTransforms.jl/pull/42/commits/2f87e1d78626a6982403f3de4ffd39eb107f4ab8)

src/transforms/replace.jl
eliascarv: Made in [2f87e1d](https://github.com/JuliaML/TableTransforms.jl/pull/42/commits/2f87e1d78626a6982403f3de4ffd39eb107f4ab8)

src/transforms/replace.jl
eliascarv: Made in [2f87e1d](https://github.com/JuliaML/TableTransforms.jl/pull/42/commits/2f87e1d78626a6982403f3de4ffd39eb107f4ab8)

src/transforms/replace.jl
```diff
@@ -0,0 +1,30 @@
...
+  olds = keys(transform.pairs)
+  inds = [findall(v -> v === old, x) .=> old for old in olds]
+  Dict(reduce(vcat, inds))
+end
+
+colapply(transform::Replace, x, c) =
+  [get(transform.pairs, xᵢ, xᵢ) for xᵢ in x]
+
+colrevert(::Replace, x, c) =
+  [get(c, i, x[i]) for i in 1:length(x)]
```
juliohm, line 30:
```suggestion
colapply(transform::Replace, x, c) = [get(transform.pairs, xᵢ, xᵢ) for xᵢ in x]

colrevert(transform::Replace, x, c) = [get(c, i, x[i]) for i in 1:length(x)]
```

src/transforms/replace.jl
eliascarv: Which of these options do you prefer:
One line:
```julia
colapply(transform::Replace, x, c) = [get(transform.pairs, xᵢ, xᵢ) for xᵢ in x]
```
Two lines:
```julia
colapply(transform::Replace, x, c) = 
  [get(transform.pairs, xᵢ, xᵢ) for xᵢ in x]
```

juliohm (approved):
The IdDict trick was clever 💯

src/transforms/replace.jl
eliascarv: Which of these options do you prefer:
One line:
```julia
colrevert(::Replace, x, c) = [get(c, i, x[i]) for i in 1:length(x)]
```
Two lines:
```julia
colrevert(::Replace, x, c) = 
  [get(c, i, x[i]) for i in 1:length(x)]
```

test/transforms.jl
```diff
@@ -672,7 +755,7 @@
     @test ntypes[6] == Int
     @test ttypes == Tables.schema(tₒ).types    
   end
-
+
```
juliohm, line 758:
```suggestion

```

eliascarv:
Thanks @juliohm!

eliascarv:
Can I merge when tests pass @juliohm?

juliohm:
Yes, feel free to merge when all green tickets show up. 👍🏽

PR #50, opened 2022-04-23
Update docstrings

ceferisbarov:
Docstrings for the following transforms were updated:
- Quantile
- Functional
- EigenAnalysis / PCA / DRS / SDS
- Sequential
- Parallel


src/transforms/functional.jl
```diff
@@ -6,6 +6,18 @@
...
+
+```julia
+Functional(cos)
+Functional(sin)
+```
+
+## Notes
+
+* The schema of the table is preserved by the transform.
+
```
juliohm, line 20: Please remove these notes. We already had this discussion in other PRs. This information is not useful.

src/transforms/quantile.jl
```diff
@@ -6,6 +6,17 @@
...
+# Examples
+
+```julia
+Quantile()
+Quantile(distribution=Normal())
+```
+
+## Notes
+
+* The schema of the table is preserved by the transform.
```
juliohm, line 19: Remove.

juliohm (changes requested):

juliohm (approved):

PR #52, opened 2022-04-24
Initial implementation of documentation

eliascarv:
~~I was inspired by the [Meshes.jl](https://juliageometry.github.io/Meshes.jl/dev/) reference guide to make this reference guide. I think the sections could have better separation and better names.~~
I'm waiting for your suggestions @juliohm.
Edit:
I put the README text in index.md and added the Transforms docstrings in transforms.md.


docs/src/traits.md
```diff
@@ -0,0 +1,18 @@
+# Traits
+
+```@docs
+Transform
+assertions
+Stateless
+Colwise
+colapply
+colrevert
+TableTransforms.colcache
```
juliohm, line 10: Maybe we need to export colcache as well?

juliohm (commented):
Can we have a preview of the docs in the PR? I remember it is possible to configure Documenter.jl and the repo to show a preview. Someone asked this recently on Zulip as well. Can you double check?

docs/src/traits.md
eliascarv: I think so. But I don't know if it's interesting to export these internal functions.

juliohm: Maybe we should unexport colapply and colrevert. What do you think?

eliascarv: I think all these functions and types from the Trait section are part of the internal API. I don't think they need to be exported.

juliohm: That is a good point. Maybe a separate PR fixing the exports first before we can finish this one here about docs.

docs/src/links.md
```diff
@@ -0,0 +1,15 @@
...
+
+```@index
+Order = [:type]
+```
+
+## Functions
+
+```@index
+Order = [:function]
+```
```
juliohm, line 15: I think we can erase these automated sections in favor of the reference guide which is already organized in sections.

docs/src/index.md
```diff
@@ -2,13 +2,228 @@
...
+a = [2randn(N÷2) .+ 6; randn(N÷2)]
+b = [3randn(N÷2); 2randn(N÷2)]
+c = randn(N)
+d = c .+ 0.6randn(N)
+table = (;a, b, c, d)
+
+# corner plot of original table
+table |> corner
+```
+![original](assets/original.png)
```
juliohm, line 78: Examples should be written using Documenter.jl example blocks. The code will run and produce the output in place. See the Meshes.jl and GeoStats.jl docs for examples.

docs/src/index.md
```diff
@@ -2,13 +2,228 @@
...
+
+```julia
+# apply transform and save cache to revert later
+newtable, cache = apply(pipeline, table)
+
+# perform additional modeling with newtable
+# newtable = ...
+
+# revert pipeline when done with modeling
+original = revert(pipeline, newtable, cache)
```
juliohm, line 120: We need to replace this example with an actual runnable code.

docs/src/index.md
```diff
@@ -2,13 +2,228 @@
...
+"fitted" with training data to unseen test data. In this case, the
+cache from a previous `apply` call is used:
+
+```julia
+# ZScore transform "fits" μ and σ using training data
+newtable, cache = apply(ZScore(), traintable)
+
+# we can reuse the same values of μ and σ with test data
+newtable = reapply(ZScore(), testtable, cache)
 ```
```
juliohm, line 133: Same here, we need a MWE.

docs/src/index.md
```diff
@@ -2,13 +2,228 @@
...
+
+## Available transforms
```
juliohm, line 135: We will have separate pages in the documentation for this section and all remaining sections. Can you create a separate page with a list of docstrings instead? We can then create a header for each transform so that users can quickly click on the name to jump to the docstring.

docs/src/transforms.md
```diff
@@ -0,0 +1,39 @@
+# Tranforms
+
+## Functions to use transforms
+
+```@docs
+apply
+revert
+isrevertible
+reapply
+```
```
juliohm, line 10: Thinking about it again, I think these should show up in the index.md along with the examples.

docs/src/transforms.md
```diff
@@ -0,0 +1,39 @@
...
+Interquartile
+ZScore
+Quantile
+Functional
+EigenAnalysis
+PCA
+DRS
+SDS
+Sequential
+→
```
juliohm, line 36: \to and \sqcup should also appear in the index.md

juliohm (changes requested):

docs/src/index.md
```diff
@@ -2,13 +2,179 @@
...
+pipeline = (f1 → f2 → f3) ⊔ (f4 → f5)
+
+# feed data into the pipeline
+table |> pipeline |> corner
 ```
 
-```@autodocs
-Modules = [TableTransforms]
+To create sequential pipeline you can use `→` (`\to<tab>`) operator and
+to crate parallel pipeline you can use `⊔` (`\sqcup<tab>`) operator:
```
juliohm, line 110:
```suggestion
Each branch is a sequence of transforms constructed with the `→` (`\to<tab>`) operator.
The branches are placed in parallel with the `⊔` (`\sqcup<tab>`) operator.
```

docs/src/index.md
```diff
@@ -2,13 +2,179 @@
...
+
+```@docs
+→
+⊔
+```
+
+### Apply and Revert
+
```
juliohm, line 118:
```suggestion
```

docs/src/index.md
```diff
@@ -2,13 +2,179 @@
...
+To revert a pipeline or single transform, use the `apply` and `revert`
```
juliohm, line 119:
```suggestion
To revert a pipeline or single transform, use the [`apply`](@ref) and [`revert`](@ref)
```

docs/src/index.md
```diff
@@ -2,13 +2,179 @@
...
+functions instead:
```
juliohm, line 120:
```suggestion
functions instead. The function [`isrevertible`](@ref) can be used to check if a transform is revertible.
```

docs/src/index.md
```diff
@@ -2,13 +2,179 @@
...
+```
+
+\
+Using the cache we can reverse the transform:
+
+```@example usage2
+original = revert(transform, newtable, cache)
+```
+
+\
```
juliohm, line 163:
```suggestion
```

docs/src/index.md
```diff
@@ -2,13 +2,179 @@
...
+
+```@docs
+apply
+revert
+isrevertible
+```
+
+To exemplify the use of these functions, let's create a table:
+
+```@example usage2
```
juliohm, line 130: No need to create a separate usage2. Just reuse the imports from usage, it will clean up the example. Also, I think we can drop the DataFrames.jl dependency from the docs and use small named tuples to clarify the idea.

docs/src/index.md
```diff
@@ -2,13 +2,179 @@
...
+using DataFrames
+using Random; Random.seed!(2) # hide
+
+a = rand(6)
+b = 3rand(6)
+c = 2rand(6)
+table = DataFrame(; a, b, c)
+```
+
+\
```
juliohm, line 141: Why backslashes? Are they necessary?

docs/src/index.md
```diff
@@ -2,13 +2,179 @@
...
+Finally, it is sometimes useful to `reapply` a transform that was
+"fitted" with training data to unseen test data. In this case, the
+cache from a previous `apply` call is used:
+
+```@docs
+reapply
+```
+
+Example:
```
juliohm, line 172:
```suggestion
Consider the following example:
```

juliohm (changes requested):

docs/src/index.md
eliascarv: These backslashes are used to create a blank line. I put them because the tables were too close to the text.

docs/src/index.md
eliascarv: I used DataFrames because it has an HTML view of the table, but I can use named tuples and PrettyTables.jl instead.

juliohm: I think a small named tuple like (a=[1,2,3], b=[4,5,6], c=[7,8,9]) or similar will be enough, no need for fancy printing. We just need to illustrate the reversibility and readers can do the checks quickly with these small number of values.

docs/src/index.md
juliohm: Let's erase the backslashes and use the default format of documenter.jl, these tables can also go away if we use small named tuples to exemplify the idea.

docs/src/transforms.md
```diff
@@ -0,0 +1,89 @@
+# Tranforms
+
+## Column selection
```
juliohm, line 3: I think we can simply use a section per transform to avoid this naming issue. It is cumbersome to type it all manually, but if you use a fancy editor like VIM or Emacs you can automate this process with a macro. Or write a Julia program that spits out the text of this documentation page.

docs/src/index.md
eliascarv: Without the HTML views of the tables, these backslashes will be unnecessary. I will remove them.

docs/src/transforms.md
juliohm: We could have used documenter.jl autodocs features but it doesn`t create a clickable index as far as I know. If you find something in their docs that allows us to automatically add all docstrings of transforms and get links for free in the tab, that would be great.

eliascarv: In this case, will the name of the section be the name of the transformation?

juliohm: Yes, too many sections but I think it is a reasonable solution for the clickable docs.

docs/src/index.md
```diff
@@ -2,13 +2,170 @@
...
+newtable
+```
+
+Using the cache we can reverse the transform:
+
+```@example usage
+original = revert(transform, newtable, cache)
+```
+
+Finally, it is sometimes useful to `reapply` a transform that was
```
juliohm, line 155:
```suggestion
Finally, it is sometimes useful to [`reapply`](@ref) a transform that was
```

juliohm (changes requested):

eliascarv:
Thanks

PR #57, opened 2022-04-28
Unexport internal types and functions; Remove unnecessary example

### LINKED ISSUE #53: Unexport internal types and functions
This types:
```julia
Transform, Stateless, Colwise
```
and this functions:
```julia
assertions, colapply, colrevert
```
are part of the internal API and should not be exported.

eliascarv:
closes #53

### DIFF
--- a/src/transforms/identity.jl
+++ b/src/transforms/identity.jl
@@ -6,12 +6,6 @@
     Identity()
 
 The identity transform `x -> x`.
-
-# Examples
-
-```julia
-Identity()
-```
 """
 struct Identity <: Stateless end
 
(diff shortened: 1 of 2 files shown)


src/transforms/identity.jl
```diff
@@ -6,12 +6,6 @@
     Identity()
 
 The identity transform `x -> x`.
-
-# Examples
-
-```julia
-Identity()
-```
```
juliohm, line 14: Is this intended? The PR is about removing exports right?

eliascarv: It's such a small change that I thought it could be included in this PR.
Want me to do a PR for this change?

juliohm: No problem, just wanted to double check it was an intentional change. 👍🏽

PR #65, opened 2022-04-30
Add RowTable and ColTable

### LINKED ISSUE #62: Introduce RowTable and ColTable transforms
For row table inputs, will the output of a transform necessarily have `rowaccess`?  Is there a general principle guiding the form of the output in general? Can I arrange that the ouput is `Tables.materialize`d? 

In MLJ  we generally `materialize` based on input type, but is has been argued that an option to specify the output type would be useful. I'm wondering what the authors of  TableTransforms.jl think about such questions.

ceferisbarov:
Fix #62 

Added RowTable transform and its tests.


src/transforms/rowtable.jl
```diff
@@ -0,0 +1,20 @@
...
+"""
+    RowTable()
+
+The transform that applies the function `Tables.rowtable` to to the input table.
+
+# Examples
+
+```julia
+RowTable()
+```
```
juliohm, line 14:
```suggestion
```

src/transforms/rowtable.jl
```diff
@@ -0,0 +1,20 @@
...
+"""
+struct RowTable <: Transform end
```
juliohm, line 16: This transform should be Stateless

test/transforms.jl
```diff
@@ -859,6 +859,19 @@
...
+  @testset "RowTable" begin
+    a = [3, 2, 1, 4, 5, 3]
+    b = [1, 4, 4, 5, 8, 5]
+    c = [1, 1, 6, 2, 4, 1]
+    t = Table(; a, b, c)
+    T = RowTable()
+    n, c = apply(T, t)
+    tₒ = revert(T, n, c)
+    @test typeof(t) <: Table
+    @test typeof(n) <: Vector
```
juliohm, line 871: Should also check that the new table is Tables.rowaccess

juliohm (changes requested):
Added a first round of reviews. Feel free to add ColTable to this PR as well given that it is pretty much the same code.

test/transforms.jl
eliascarv: I think this check is not necessary:
```julia
@test typeof(t) <: Table
```

src/transforms/coltable.jl
```diff
@@ -0,0 +1,14 @@
...
+"""
+    ColTable()
+
+The transform that applies the function `Tables.columntable` to to the input table.
+"""
+struct ColTable <: Stateless end
+
+apply(::ColTable, table) = Tables.columntable(table), table
+
+revert(::ColTable, table, cache) = cache
```
eliascarv, line 14:
```suggestion
revert(::ColTable, newtable, cache) = cache
```

src/transforms/rowtable.jl
```diff
@@ -0,0 +1,14 @@
...
+"""
+    RowTable()
+
+The transform that applies the function `Tables.rowtable` to to the input table.
+"""
+struct RowTable <: Stateless end
+
+apply(::RowTable, table) = Tables.rowtable(table), table
+
+revert(::RowTable, table, cache) = cache
```
eliascarv, line 14:
```suggestion
revert(::RowTable, newtable, cache) = cache
```

src/transforms/coltable.jl
juliohm: @ceferisbarov please pay more attention to the code style. Try to copy/paste the variable names that are used consistently across the transforms. Thank you @eliascarv for spotting that.

eliascarv (approved):

juliohm (changes requested):
Awesome, now we need to add the two transforms to the documentation like the other transforms.

test/transforms.jl
```diff
@@ -859,6 +859,32 @@
...
+  @testset "RowTable" begin
+    a = [3, 2, 1, 4, 5, 3]
+    b = [1, 4, 4, 5, 8, 5]
+    c = [1, 1, 6, 2, 4, 1]
+    t = Table(; a, b, c)
+    T = RowTable()
+    n, c = apply(T, t)
+    tₒ = revert(T, n, c)
+    @test typeof(n) <: Vector
+    @test Tables.rowaccess(typeof(n))
```
juliohm, line 871: Do we really need the typeof here @ceferisbarov ?

test/transforms.jl
```diff
@@ -859,6 +859,32 @@
...
+  @testset "ColTable" begin
+    a = [3, 2, 1, 4, 5, 3]
+    b = [1, 4, 4, 5, 8, 5]
+    c = [1, 1, 6, 2, 4, 1]
+    t = Table(; a, b, c)
+    T = ColTable()
+    n, c = apply(T, t)
+    tₒ = revert(T, n, c)
+    @test typeof(n) <: NamedTuple
+    @test Tables.columnaccess(typeof(n))
```
juliohm, line 884: Same here, do we need typeof?

test/transforms.jl
ceferisbarov: From Tables.jl API:  
_

Tables.rowaccess(table)    Declare that your table type defines a Tables.rows(table) method_  
  
This seems to be the intended use, but it also works when I remove `typeof`.

juliohm:
Awesome! Thank you! 💯

PR #68, opened 2022-05-05
Add StdNames

ceferisbarov:
I am still trying to make the code more readable. Suggestions are welcome.

### DIFF
--- a/Project.toml
+++ b/Project.toml
@@ -1,7 +1,7 @@
 name = "TableTransforms"
 uuid = "0d432bfd-3ee1-4ac1-886a-39f05cc69a3e"
 authors = ["Júlio Hoffimann <julio.hoffimann@gmail.com> and contributors"]
-version = "0.5.2"
+version = "0.2.0"
 
 [deps]
 Distributions = "31c24e10-a181-5473-b8eb-7969acd0382f"
--- a/Project.toml
+++ b/Project.toml
@@ -1,7 +1,7 @@
 name = "TableTransforms"
 uuid = "0d432bfd-3ee1-4ac1-886a-39f05cc69a3e"
 authors = ["Júlio Hoffimann <julio.hoffimann@gmail.com> and contributors"]
-version = "0.2.0"
+version = "1.0.0"
 
 [deps]
 Distributions = "31c24e10-a181-5473-b8eb-7969acd0382f"
--- a/src/transforms/stdnames.jl
+++ b/src/transforms/stdnames.jl
@@ -5,15 +5,19 @@
 """
     StdNames(:spec)
 
-The transform that standardizes the column names. There are three different methods of standardization: uppercase, camelcase, and snakecase.
-They can be specified by the `spec`` argument (`:upper`, `:camel`, and `:snake`, respectively). Transform defaults to uppercase. 
+The transform that standardizes the column names.
+There are three different methods of standardization: `:upper`, `:camel`, and `:snake`.
+Transform defaults to `:upper`.
 
 # Examples
 
-```julia
-StdNames()
-StdNames(:camel)
-```
+| Transform | Original Col. Name | Transformed Col. Name |
+| ----------- | ----------- |	----------- |
+| `StdNames`    		 | apple_Trees | APPLETREES |
+| `StdNames(:upper)` | apple_Trees | APPLETREES |
+| `StdNames(:camel)` | apple_Trees | AppleTrees |
+| `StdNames(:snake)` | apple_Trees | apple_trees |
+
 """
 struct StdNames <: Stateless
   spec::Symbol
--- a/src/transforms/stdnames.jl
+++ b/src/transforms/stdnames.jl
@@ -25,13 +25,13 @@ isrevertible(::Type{StdNames}) = true
 function apply(transform::StdNames, table)  
   cols = Tables.columns(table)
   oldnames = string.(Tables.columnnames(cols))
-  preprocessed = _filter.(oldnames)
+  cleaned = _clean.(oldnames)
   
   spec = transform.spec
 
-  spec == :camel && (names = _camel.(preprocessed))
-  spec == :snake && (names = _snake.(preprocessed))
-  spec == :upper && (names = _upper.(preprocessed))
+  spec == :camel && (names = _camel.(cleaned))
+  spec == :snake && (names = _snake.(cleaned))
+  spec == :upper && (names = _upper.(cleaned))
 
   newnames = _unique(names)
   oldnew = Dict(zip(oldnames, newnames))
--- a/src/transforms/stdnames.jl
+++ b/src/transforms/stdnames.jl
@@ -1,0 +1,59 @@
+# ------------------------------------------------------------------
+# Licensed under the MIT License. See LICENSE in the project root.
+# ------------------------------------------------------------------
+
+"""
+    StdNames(S)
+
+StdNames transform standardizes the column names.
+"""
+struct StdNames <: Stateless
+    stdmethod::Symbol
+end
+
+isrevertible(::Type{StdNames}) = true
+
+function apply(transform::StdNames, table)    
+    if transform.stdmethod == :camel
+        _camel(table)
+    elseif transform.stdmethod == :snake
+        _snake(table)
+    else
+        _upper(table)
+    end
+end
+
+function revert(::StdNames, newtable, cache)
+    names = Tables.columnnames(newtable)
+    namesdict = Dict(zip(names, cache))
+    newtable |> Rename(namesdict)
+end
+
+function _camel(table)
+    oldnames = Tables.columnnames(table)
+    function camelfunction(x)
+        substrings = split(string(x))
+        capitalize(s) = String([i == 1 ? uppercase(c) : c for (i, c) in enumerate(s)])
+        Symbol(join(map(capitalize, substrings)))
+    end
+
+    newnames = map(camelfunction, oldnames)
+    namesdict = Dict(zip(oldnames, newnames))
+    table |> Rename(namesdict), oldnames
+end
+
+function _snake(table)
+    oldnames = Tables.columnnames(table)
+    snakefunction(x) = Symbol(join(split(string(x)), "_"))
+    newnames = map(snakefunction, oldnames)
+    namesdict = Dict(zip(oldnames, newnames))
+    table |> Rename(namesdict), oldnames
+end
+
+function _upper(table)
+    oldnames = Tables.columnnames(table)
+    upperfunction(x) = Symbol(uppercase(string(x)))    
+    newnames = map(upperfunction, oldnames)
+    namesdict = Dict(zip(oldnames, newnames))
+    table |> Rename(namesdict), oldnames
+end
--- a/src/transforms/stdnames.jl
+++ b/src/transforms/stdnames.jl
@@ -38,10 +38,11 @@ function apply(transform::StdNames, table)
   spec == :snake && (names = _snake.(cleaned))
   spec == :upper && (names = _upper.(cleaned))
 
+  # make names unique
   newnames = _unique(names)
-  oldnew = Dict(zip(oldnames, newnames))
-
-  rtrans = Rename(oldnew)
+  
+  # rename transform
+  rtrans = Rename(Dict(oldnames .=> newnames))
   newtable, rcache = apply(rtrans, table)
 
   newtable, (rtrans, rcache)
--- a/src/transforms/stdnames.jl
+++ b/src/transforms/stdnames.jl
@@ -21,14 +21,14 @@ function apply(transform::StdNames, table)
   
   spec = transform.spec
 
-  spec == :camel && (nonuniquenames = _camel.(oldnames))
-  spec == :snake && (nonuniquenames = _snake.(oldnames))
-  spec == :upper && (nonuniquenames = _upper.(oldnames))
+  spec == :camel && (names = _camel.(oldnames))
+  spec == :snake && (names = _snake.(oldnames))
+  spec == :upper && (names = _upper.(oldnames))
 
-  newnames = _unique(nonuniquenames)
-  names = Dict(zip(oldnames, newnames))
+  newnames = _unique(names)
+  oldnew = Dict(zip(oldnames, newnames))
 
-  rtrans = Rename(names)
+  rtrans = Rename(oldnew)
   newtable, rcache = apply(rtrans, table)
 
   newtable, (rtrans, rcache)
--- a/src/transforms/stdnames.jl
+++ b/src/transforms/stdnames.jl
@@ -5,7 +5,15 @@
 """
     StdNames(:spec)
 
-StdNames transform standardizes the column names.
+The transform that standardizes the column names. There are three different methods of standardization: uppercase, camelcase, and snakecase.
+They can be specified by the `spec`` argument (`:upper`, `:camel`, and `:snake`, respectively). Transform defaults to uppercase. 
+
+# Examples
+
+```julia
+StdNames()
+StdNames(:camel)
+```
 """
 struct StdNames <: Stateless
   spec::Symbol
@@ -56,6 +64,6 @@ end
 
 _camel(name) = join(uppercasefirst.(split(strip(name, delim), delim)))
 
-_snake(name) = lowercase(join(split(strip(name, delim), delim), '_'))
+_snake(name) = join(lowercase.(split(strip(name, delim), delim)), '_')
 
-_upper(name) = replace(uppercase(strip(name, delim)), delim => "")
+_upper(name) = replace(uppercase(name), delim => "")
--- a/src/transforms/stdnames.jl
+++ b/src/transforms/stdnames.jl
@@ -45,5 +45,5 @@ function _snake(name)
 end
 
 function _upper(name)
-  Symbol(replace(uppercase(string(name)), " "=>""))
+  Symbol(replace(uppercase(string(name)), " " => ""))
 end
--- a/src/transforms/stdnames.jl
+++ b/src/transforms/stdnames.jl
@@ -17,16 +17,15 @@ isrevertible(::Type{StdNames}) = true
 
 function apply(transform::StdNames, table)  
   cols = Tables.columns(table)
-  oldnames = Tables.columnnames(cols)
+  oldnames = string.(Tables.columnnames(cols))
   
   spec = transform.spec
 
-  spec == :camel && (newstringnames = _camel.(string.(oldnames)))
-  spec == :snake && (newstringnames = _snake.(string.(oldnames)))
-  spec == :upper && (newstringnames = _upper.(string.(oldnames)))
+  spec == :camel && (nonuniquenames = _camel.(oldnames))
+  spec == :snake && (nonuniquenames = _snake.(oldnames))
+  spec == :upper && (nonuniquenames = _upper.(oldnames))
 
-  newuniquenames = _unique(newstringnames)
-  newnames = Symbol.(newuniquenames)
+  newnames = _unique(nonuniquenames)
   names = Dict(zip(oldnames, newnames))
 
   rtrans = Rename(names)
@@ -43,15 +42,16 @@ end
 const delim = ['_', ' ']
 
 function _unique(names)
-  uniquenames = [names...]
-  for i in range(2,length(uniquenames))
-      prev = uniquenames[1:i-1]
-      while uniquenames[i] in prev
-        uniquenames[i] = string(uniquenames[i], "_")
-      end
+  newnames = Vector{String}()
+  for name in names
+    updatedname = name
+    while updatedname ∈ newnames
+      updatedname = string(updatedname, "_")
+    end
+    push!(newnames, updatedname)
   end
 
-  Tuple(uniquenames)
+  newnames
 end
 
 _camel(name) = join(uppercasefirst.(split(strip(name, delim), delim)))
--- a/src/transforms/stdnames.jl
+++ b/src/transforms/stdnames.jl
@@ -16,7 +16,8 @@ StdNames() = StdNames(:upper)
 isrevertible(::Type{StdNames}) = true
 
 function apply(transform::StdNames, table)  
-  oldnames = Tables.columnnames(table)
+  cols = Tables.columns(table)
+  oldnames = Tables.columnnames(cols)
   spec = transform.spec
 
   (spec == :camel) && (newnames = map(_camel, oldnames))
@@ -24,25 +25,25 @@ function apply(transform::StdNames, table)
   (spec == :upper) && (newnames = map(_upper, oldnames))
 
   names = Dict(zip(oldnames, newnames))
-  table |> Rename(names), oldnames
+  rtrans = Rename(names)
+  newtable, rcache = apply(rtrans, table)
+  newtable, (rtrans, rcache)
 end
 
 function revert(::StdNames, newtable, cache)
-  newnames = Tables.columnnames(newtable)
-  names = Dict(zip(newnames, cache))
-  newtable |> Rename(names)
+  rtrans, rcache = cache
+  revert(rtrans, newtable, rcache)
 end
 
 function _camel(name)
   substrings = split(string(name))
-  capitalize(s) = String([i == 1 ? uppercase(c) : c for (i, c) in enumerate(s)])
-  Symbol(join(map(capitalize, substrings)))
+  Symbol(join(map(uppercasefirst, substrings)))
 end
 
 function _snake(name)
-  Symbol(join(split(string(name)), "_"))
+  Symbol(lowercase(join(split(string(name)), "_")))
 end
 
 function _upper(name)
-  Symbol(uppercase(string(name)))
+  Symbol(replace(uppercase(string(name)), " "=>""))
 end
--- a/src/transforms/stdnames.jl
+++ b/src/transforms/stdnames.jl
@@ -20,10 +20,11 @@ function apply(transform::StdNames, table)
   oldnames = Tables.columnnames(cols)
   spec = transform.spec
 
-  (spec == :camel) && (newnames = map(_camel, oldnames))
-  (spec == :snake) && (newnames = map(_snake, oldnames))
-  (spec == :upper) && (newnames = map(_upper, oldnames))
+  (spec == :camel) && (newnames = map(_camel, map(string, oldnames)))
+  (spec == :snake) && (newnames = map(_snake, map(string, oldnames)))
+  (spec == :upper) && (newnames = map(_upper, map(string, oldnames)))
 
+  newnames = map(Symbol, newnames)
   names = Dict(zip(oldnames, newnames))
   rtrans = Rename(names)
   newtable, rcache = apply(rtrans, table)
@@ -36,14 +37,14 @@ function revert(::StdNames, newtable, cache)
 end
 
 function _camel(name)
-  substrings = split(string(name))
-  Symbol(join(map(uppercasefirst, substrings)))
+  substrings = split(name)
+  join(map(uppercasefirst, substrings))
 end
 
 function _snake(name)
-  Symbol(lowercase(join(split(string(name)), "_")))
+  lowercase(join(split(name), "_"))
 end
 
 function _upper(name)
-  Symbol(replace(uppercase(string(name)), " " => ""))
+  replace(uppercase(name), " " => "")
 end
--- a/src/transforms/stdnames.jl
+++ b/src/transforms/stdnames.jl
@@ -66,7 +66,7 @@ function _unique(names)
   newnames
 end
 
-_camel(name) = join(uppercasefirst.(split(strip(name, delim), delim)))
+_camel(name) = join(uppercasefirst.(split(name, delim)))
 
 _snake(name) = join(lowercase.(split(strip(name, delim), delim)), '_')
 
--- a/src/transforms/stdnames.jl
+++ b/src/transforms/stdnames.jl
@@ -21,11 +21,12 @@ function apply(transform::StdNames, table)
   
   spec = transform.spec
 
-  (spec == :camel) && (newstringnames = _camel.(string.(oldnames)))
-  (spec == :snake) && (newstringnames = _snake.(string.(oldnames)))
-  (spec == :upper) && (newstringnames = _upper.(string.(oldnames)))
+  spec == :camel && (newstringnames = _camel.(string.(oldnames)))
+  spec == :snake && (newstringnames = _snake.(string.(oldnames)))
+  spec == :upper && (newstringnames = _upper.(string.(oldnames)))
 
-  newnames = Symbol.(newstringnames)
+  newuniquenames = _unique(newstringnames)
+  newnames = Symbol.(newuniquenames)
   names = Dict(zip(oldnames, newnames))
 
   rtrans = Rename(names)
@@ -39,8 +40,22 @@ function revert(::StdNames, newtable, cache)
   revert(rtrans, newtable, rcache)
 end
 
-_camel(name) = join(uppercasefirst.(split(strip(name, ['_', ' ']), ['_', ' '])))
+const delim = ['_', ' ']
 
-_snake(name) = lowercase(join(split(strip(name, ['_', ' ']), ['_', ' ']), '_'))
+function _unique(names)
+  uniquenames = [names...]
+  for i in range(2,length(uniquenames))
+      prev = uniquenames[1:i-1]
+      while uniquenames[i] in prev
+        uniquenames[i] = string(uniquenames[i], "_")
+      end
+  end
 
-_upper(name) = replace(uppercase(strip(name, ['_', ' '])), " " => "", "_" => "")
+  Tuple(uniquenames)
+end
+
+_camel(name) = join(uppercasefirst.(split(strip(name, delim), delim)))
+
+_snake(name) = lowercase(join(split(strip(name, delim), delim), '_'))
+
+_upper(name) = replace(uppercase(strip(name, delim)), delim => "")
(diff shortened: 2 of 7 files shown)


src/transforms/stdnames.jl
```diff
@@ -0,0 +1,59 @@
...
+# Licensed under the MIT License. See LICENSE in the project root.
+# ------------------------------------------------------------------
+
+"""
+    StdNames(S)
+
+StdNames transform standardizes the column names.
+"""
+struct StdNames <: Stateless
+    stdmethod::Symbol
```
juliohm, line 11: We use 2-space indentation

juliohm, line 11: rename the field to `spec`

src/transforms/stdnames.jl
```diff
@@ -0,0 +1,59 @@
...
+isrevertible(::Type{StdNames}) = true
+
+function apply(transform::StdNames, table)    
+    if transform.stdmethod == :camel
+        _camel(table)
+    elseif transform.stdmethod == :snake
+        _snake(table)
+    else
+        _upper(table)
+    end
```
juliohm, line 23: You can use the short cut:

```julia
spec = transform.spec

spec == :camel && _camel(table)
spec == :snake && _snake(table)
spec == :upper && _upper(table)
```

src/transforms/stdnames.jl
```diff
@@ -0,0 +1,59 @@
...
+end
+
+function revert(::StdNames, newtable, cache)
+    names = Tables.columnnames(newtable)
+    namesdict = Dict(zip(names, cache))
+    newtable |> Rename(namesdict)
```
juliohm, line 29: Try to use the same variable names used in other transforms.

src/transforms/stdnames.jl
```diff
@@ -0,0 +1,59 @@
...
+    table |> Rename(namesdict), oldnames
+end
+
+function _upper(table)
+    oldnames = Tables.columnnames(table)
+    upperfunction(x) = Symbol(uppercase(string(x)))    
+    newnames = map(upperfunction, oldnames)
+    namesdict = Dict(zip(oldnames, newnames))
+    table |> Rename(namesdict), oldnames
+end
```
juliohm, line 59: Notice how these three functions are repeating the same code and logic. Try to move the common pieces to the `apply` function and then only call specific _camel(str) on the names. In other words, your `apply` should contain all logic and loop over all names performing the standardization with one of three functions _camel, _snake and _upper, which should operate on strings.

test/transforms.jl
```diff
@@ -680,6 +680,34 @@
     @test n1 == n2
   end
 
+  @testset "StdNames" begin
```
juliohm, line 683: Make sure you include tests with row tables. Take a look at other transforms for examples.

juliohm (changes requested):
Great start @ceferisbarov , left a first round of reviews.

src/transforms/stdnames.jl
ceferisbarov: This doesn't work as expected. It works fine while `spec = :upper`, but it returns `false` otherwise.

juliohm: I may have missed parenthesis. What about `(spec == :camel) && _camel(table)`? Anyways, I think you will need to refactor the logic a bit more to move more lines of code to apply and refactor _camel, _snake and _upper to take strings as input.

test/transforms.jl
```diff
@@ -680,6 +680,34 @@
...
+    a = [3, 2, 1, 4, 5, 3]
+    b = [2, 4, 4, 5, 8, 5]
+    t = Table(; a, b)
+    t = t |> Rename(:a => "apple tree", :b => "banana tree")
```
juliohm, line 687: You can create a table with the desired column names instead of using Rename here.

juliohm: Also, in this example make sure that the table also contains clean column names such as `a`, `B`. They shouldn't be affected by some of the conventions.

ceferisbarov: @juliohm  I have tried using a Symbol or a string instead of a variable name, but both of them throw syntax error. I am not sure how I can create a table with these column names.

src/transforms/stdnames.jl
ceferisbarov: I assumed you were referring to `namesdict` variable.

src/transforms/stdnames.jl
```diff
@@ -0,0 +1,48 @@
...
+function apply(transform::StdNames, table)  
+  oldnames = Tables.columnnames(table)
+  spec = transform.spec
+
+  (spec == :camel) && (newnames = map(_camel, oldnames))
+  (spec == :snake) && (newnames = map(_snake, oldnames))
+  (spec == :upper) && (newnames = map(_upper, oldnames))
+
+  names = Dict(zip(oldnames, newnames))
+  table |> Rename(names), oldnames
```
juliohm, line 27: Now that is 🥇 👏🏽

src/transforms/stdnames.jl
```diff
@@ -0,0 +1,48 @@
...
+end
+
+function _camel(name)
+  substrings = split(string(name))
+  capitalize(s) = String([i == 1 ? uppercase(c) : c for (i, c) in enumerate(s)])
+  Symbol(join(map(capitalize, substrings)))
+end
+
+function _snake(name)
+  Symbol(join(split(string(name)), "_"))
```
juliohm, line 43: You also need to change everything to `lowercase`

src/transforms/stdnames.jl
```diff
@@ -0,0 +1,48 @@
...
+end
+
+function _upper(name)
+  Symbol(uppercase(string(name)))
```
juliohm, line 47: You also need to remove whitespace

test/transforms.jl
```diff
@@ -680,6 +680,49 @@
     @test n1 == n2
   end
 
+  @testset "StdNames" begin
+    names = (:a, Symbol("apple tree"), Symbol("banana tree"))
```
juliohm, line 684: I suggest adding more complicated names such as `foo_Bar`, `A b`, ... you need to test the names extensively.

Ideally you will introduce a small table for each kind of name issue. For example, you can introduce a table with names that have underscore, another table with names that have spaces, another table with upper and lower case mixed, etc. This will cover more issues.

juliohm (changes requested):

src/transforms/stdnames.jl
juliohm: We can do even better here and save the Rename transform as the cache, so that the revert is just a single call to revert with the cache transform.

src/transforms/stdnames.jl
```diff
@@ -0,0 +1,48 @@
...
+struct StdNames <: Stateless
+  spec::Symbol
+end
+
+StdNames() = StdNames(:upper)
+
+isrevertible(::Type{StdNames}) = true
+
+function apply(transform::StdNames, table)  
+  oldnames = Tables.columnnames(table)
```
eliascarv, line 19:
```suggestion
  cols = Tables.columns(table)
  oldnames = Tables.columnnames(cols)
```
This is the correct way to use the `Tables.columnnames` function.

eliascarv (changes requested):

src/transforms/stdnames.jl
ceferisbarov: Thanks!

src/transforms/stdnames.jl
ceferisbarov: I did this, but it looks a little bit off. `revert` function of the `Rename` transform takes three arguments `(transform, table, cache)`, but only uses two `(transform, table)`. So we get something like this:
```
revert(cache, newtable, cache)
```
The first cache is cache of StdNames transform. The second cache is supposes to be cache of the Rename transform, but it is not used inside the function anyway, so I just wrote it like this. Wait until I push the commit. It will be cleared then.

eliascarv: I think you are not understanding. This is what Júlio is trying to say:
```julia
function apply(transform::StdNames, table)
  # code...
  names = Dict(zip(oldnames, newnames))
  rtrans = Revert(names)
  newtable, cache = apply(rtrans, table)
  newtable, (rtrans, rcache)
end

revert(::StdNames, newtable, cache)
  rtrans, rcache = cache
  revert(rtrans, newtable, rcache)
end
```

eliascarv: I think you didn't understand what Júlio meant. This would be the code:
```julia
function apply(transform::StdNames, table)
  # code...
  names = Dict(zip(oldnames, newnames))
  rtrans = Revert(names)
  newtable, cache = apply(rtrans, table)
  newtable, (rtrans, rcache)
end

revert(::StdNames, newtable, cache)
  rtrans, rcache = cache
  revert(rtrans, newtable, rcache)
end
```

ceferisbarov: Oh, I see. Thanks again. You are doing most of the work for me here :))

test/transforms.jl
ceferisbarov: @juliohm  I know we already discussed this, but don't you think we should also split at underscores? I actually think it would be a reasonable thing to do.

ceferisbarov:
@juliohm  Test set grew too large. I think we can create a function to avoid repeating the same lines.

test/transforms.jl
juliohm: @ceferisbarov it depends on the transform, for `:snake` we don't want to split underscores.

juliohm (changes requested):
That is looking great now @ceferisbarov. I have a final suggestion before we merge this. Can you please replace all tests you have written by simple tests of the form:

```julia
@test TableTransforms._camel("apple banana") == "AppleBanana"
@test TableTransforms._snake("apple banana") == "apple_banana"
...
```

That way we can make sure that we cover various types of strings.

After the above set of tests you can simply add two more tests with tables to make sure that `apply` and `revert` are working with column and row tables. Also `reapply` and `isrevertible`.

src/transforms/stdnames.jl
```diff
@@ -0,0 +1,49 @@
...
+  substrings = split(string(name))
+  Symbol(join(map(uppercasefirst, substrings)))
+end
+
+function _snake(name)
+  Symbol(lowercase(join(split(string(name)), "_")))
+end
+
+function _upper(name)
+  Symbol(replace(uppercase(string(name)), " "=>""))
```
eliascarv, line 48:
```suggestion
  Symbol(replace(uppercase(string(name)), " " => ""))
```
Just a small code style change.

src/transforms/stdnames.jl
```diff
@@ -0,0 +1,49 @@
...
+  Symbol(join(map(uppercasefirst, substrings)))
+end
+
+function _snake(name)
+  Symbol(lowercase(join(split(string(name)), "_")))
+end
+
+function _upper(name)
+  Symbol(replace(uppercase(string(name)), " " => ""))
+end
```
juliohm, line 49: Refactor these functions so that they receive String and output String. Then in the main `apply` call you can convert from and to Symbol. This will clean up the code considerably.

src/transforms/stdnames.jl
```diff
@@ -0,0 +1,50 @@
...
+isrevertible(::Type{StdNames}) = true
+
+function apply(transform::StdNames, table)  
+  cols = Tables.columns(table)
+  oldnames = Tables.columnnames(cols)
+  spec = transform.spec
+
+  (spec == :camel) && (newnames = map(_camel, map(string, oldnames)))
+  (spec == :snake) && (newnames = map(_snake, map(string, oldnames)))
+  (spec == :upper) && (newnames = map(_upper, map(string, oldnames)))
```
juliohm, line 25: Instead of using map(f, xs) everywhere, try to use the broadcast syntax: `string.(oldnames)`

src/transforms/stdnames.jl
```diff
@@ -0,0 +1,50 @@
...
+
+  newnames = map(Symbol, newnames)
```
juliohm, line 27: You don't overwrite variables in Julia because this will lead to type instability often.

src/transforms/stdnames.jl
```diff
@@ -0,0 +1,50 @@
...
+  names = Dict(zip(oldnames, newnames))
+  rtrans = Rename(names)
+  newtable, rcache = apply(rtrans, table)
+  newtable, (rtrans, rcache)
```
juliohm, line 31: Notice that this paragraph contains different steps of the algorithm. Try to split lines into paragraphs that make a cohesive sense.

src/transforms/stdnames.jl
```diff
@@ -0,0 +1,50 @@
...
+  join(map(uppercasefirst, substrings))
+end
+
+function _snake(name)
+  lowercase(join(split(name), "_"))
+end
+
+function _upper(name)
+  replace(uppercase(name), " " => "")
+end
```
juliohm, line 50: Note the inconsistency. Sometimes you create an auxiliary variable substrings, sometimes you don't. Try to improve your consistency writing code over time. Also, these functions could be oneliners.

juliohm: Replace the `map` by dot syntax when possible.

test/transforms.jl
```diff
@@ -680,6 +680,49 @@
     @test n1 == n2
   end
 
+  @testset "StdNames" begin
+    # camel, snake, upper tests
+    @test TableTransforms._camel("apple banana") == "AppleBanana"
+    @test TableTransforms._snake("apple banana") == "apple_banana"
+    @test TableTransforms._upper("apple banana") == "APPLEBANANA"
+
+    @test TableTransforms._camel("apple_banana") == "Apple_banana"
```
juliohm, line 689: The camel case should remove the underscore and handle the second word with a first capital letter.

test/transforms.jl
```diff
@@ -680,6 +680,49 @@
...
+    @test TableTransforms._snake("apple_banana") == "apple_banana"
+    @test TableTransforms._upper("apple_banana") == "APPLE_BANANA"
```
juliohm, line 691: I think the upper case should also remove the underscore?

test/transforms.jl
```diff
@@ -680,6 +680,49 @@
...
+
+    @test TableTransforms._camel("apple_Banana") == "Apple_Banana"
+    @test TableTransforms._snake("apple_Banana") == "apple_banana"
+    @test TableTransforms._upper("apple_Banana") == "APPLE_BANANA"
```
juliohm, line 695: I think we should organize these tests differently. For example, consider the camel case. Your would like to write a list of strings like `["apple banana", "Apple_banana", ...]` and then apply the camel case function to all elements of the list to make sure all results are equal to `"AppleBanana"`.

test/transforms.jl
```diff
@@ -680,6 +680,49 @@
...
+    @test isrevertible(T)
+    rtₒ = revert(T, n, c)
+    @test rt == rtₒ
+
+    # reapply test
+    T = StdNames()
+    n1, c1 = apply(T, rt)
+    n2 = reapply(T, n1, c1)
+    @test n1 == n2
+
```
juliohm, line 723:
```suggestion
```

test/transforms.jl
```diff
@@ -680,6 +680,49 @@
...
+    @test TableTransforms._camel("B") == "B"
+    @test TableTransforms._snake("B") == "b"
+    @test TableTransforms._upper("B") == "B"
+
+    # row table test
+    names = (:a, Symbol("apple tree"), Symbol("banana tree"))
+    cols = ([1,2,3], [4,5,6], [7,8,9])
+    t = Table(; zip(names, cols)...)
+
```
juliohm, line 709: Why there is an empty line here? Can you please try to be more consistent with the way you group lines of code? It helps a lot the reviewer understand your intention.

juliohm (changes requested):
@ceferisbarov please request a review whenever you feel that the PR is ready. Attached you can find another round of suggestions.

test/transforms.jl
ceferisbarov: @juliohm I am trying something like this now, but I came across an edge case. What should `TableTransforms._snake("_A")` return? Do we strip spaces and underscores or not?

juliohm (changes requested):

src/transforms/stdnames.jl
```diff
@@ -0,0 +1,45 @@
...
+function apply(transform::StdNames, table)  
+  cols = Tables.columns(table)
+  oldnames = Tables.columnnames(cols)
+  spec = transform.spec
+
+  (spec == :camel) && (newstringnames = _camel.(string.(oldnames)))
+  (spec == :snake) && (newstringnames = _snake.(string.(oldnames)))
+  (spec == :upper) && (newstringnames = _upper.(string.(oldnames)))
+
+  newnames = Symbol.(newstringnames)
```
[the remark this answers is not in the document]
ceferisbarov: Rename struct is defined as:
```
struct Rename <: Stateless
  names::Dict{Symbol,Symbol}
end
```
And its constructors:
```
Rename(names::Pair) = _pairsyms(names) |> Dict |> Rename
Rename(names...) = _pairsyms.(names) |> Dict |> Rename

_pairsyms(x::Pair) = Symbol(first(x)) => Symbol(last(x))
```
We can pass `string => Symbol`, `Symbol => string`, and `string => string` pairs to `Rename`, and it will work. But if we pass them as a dictionary (which is what we are doing in this case), it will throw `MethodError: Cannot convert an object of type String to an object of type Symbol`.

I think this is inconsistent behavior and should be changed. We can update the `Rename` transform to accept strings as part of a dictionary, too. But that is beyond the scope of the PR. @juliohm

ceferisbarov:
@juliohm  We have another interesting edge case. Different column names can become same after transformation. For example, `apple_tree` and `appletree` would both be transformed into `APPLETREE` via `StdNames(:upper)` transform. This throws the following error:
```
duplicate field name in NamedTuple: "APPLETREES" is not unique
```
Now it obviously doesn't make sense to name two different columns `apple_tree` and `appletree`, but it is possible in theory. What do you suggest?

ceferisbarov:
@julai Sorry, I read your last comment only after writing mine :)

src/transforms/stdnames.jl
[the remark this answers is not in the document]
juliohm: Yes, we can work on that on a separate PR.

juliohm:
@ceferisbarov I think I already suggested a fix for this non-uniqueness issue. Did you read my comment above? You should apply the _camel or whatever function to all column names and then make them unique afterwards as a post-processing step before calling Rename. This post-processing step will append as many _ as necessary to make the columns unique. For example, if the names are A A A they will become A A_ A__

juliohm:
You pinged a different github account

src/transforms/stdnames.jl
```diff
@@ -0,0 +1,61 @@
...
+  uniquenames = [names...]
+  for i in range(2,length(uniquenames))
+      prev = uniquenames[1:i-1]
+      while uniquenames[i] in prev
+        uniquenames[i] = string(uniquenames[i], "_")
+      end
+  end
+
+  Tuple(uniquenames)
+end
```
juliohm, line 55: Again, you are mixing 2-space indentation with 4-space indentation. Please use an IDE and set this to avoid bumping into this issue again? Also, the implementation is not great. Simply create an empty vector of Symbol and keep pushing to it if the entry is not already in the vector. Don't use indexing.

src/transforms/stdnames.jl
```diff
@@ -0,0 +1,61 @@
...
+  
+  spec = transform.spec
+
+  spec == :camel && (newstringnames = _camel.(string.(oldnames)))
+  spec == :snake && (newstringnames = _snake.(string.(oldnames)))
+  spec == :upper && (newstringnames = _upper.(string.(oldnames)))
+
+  newuniquenames = _unique(newstringnames)
+  newnames = Symbol.(newuniquenames)
+  names = Dict(zip(oldnames, newnames))
```
juliohm, line 30: Let's first work on a PR to add a Rename with Dict of String, then we can come back here and avoid all this back and forth between Symbol and String.

juliohm (changes requested):
@ceferisbarov please start a separate PR adding a Rename with Dict of Strings. After that we can come back to this PR and finish it.

juliohm:
Now that #71 has been merged, we can come back to this PR. @ceferisbarov please rebase this PR to the latest master branch so that we can use the latest Dict constructor you added 👍🏽

src/transforms/stdnames.jl
ceferisbarov: Is this what you had in mind? I don't think our vector needs to be `Vector{Symbol}`, as we are working with strings.
```julia
function unique(names)
  newnames = Vector{String}()
  for name in names
    updatedname = name
    while updatedname ∈ newnames
      updatedname = string(updatedname, "_")
    end
    push!(newnames, updatedname)
  end

  newnames
end
```

Project.toml
```diff
@@ -1,7 +1,7 @@
 name = "TableTransforms"
 uuid = "0d432bfd-3ee1-4ac1-886a-39f05cc69a3e"
 authors = ["Júlio Hoffimann <julio.hoffimann@gmail.com> and contributors"]
-version = "0.5.2"
+version = "1.0.0"
```
juliohm, line 4: This change should not show up in this PR. I think you rebased with the incorrect commit. Rebasing is always a nightmare. Please feel free to start a fresh PR if you can't sort out the rebase.

juliohm (changes requested):

src/transforms/stdnames.jl
```diff
@@ -0,0 +1,61 @@
...
+  spec = transform.spec
+
+  spec == :camel && (nonuniquenames = _camel.(oldnames))
+  spec == :snake && (nonuniquenames = _snake.(oldnames))
+  spec == :upper && (nonuniquenames = _upper.(oldnames))
+
+  newnames = _unique(nonuniquenames)
+  names = Dict(zip(oldnames, newnames))
+
+  rtrans = Rename(names)
```
juliohm, line 31:
```suggestion
  spec == :camel && (names = _camel.(oldnames))
  spec == :snake && (names = _snake.(oldnames))
  spec == :upper && (names = _upper.(oldnames))

  newnames = _unique(names)
  oldnew = Dict(zip(oldnames, newnames))

  rtrans = Rename(oldnew)
```

src/transforms/stdnames.jl
```diff
@@ -0,0 +1,61 @@
...
+
+function revert(::StdNames, newtable, cache)
+  rtrans, rcache = cache
+  revert(rtrans, newtable, rcache)
+end
+
+const delim = ['_', ' ']
+
+function _unique(names)
+  newnames = Vector{String}()
```
juliohm, line 45:
```suggestion
  newnames = String[]
```

src/transforms/stdnames.jl
```diff
@@ -0,0 +1,61 @@
...
+  for name in names
+    updatedname = name
+    while updatedname ∈ newnames
+      updatedname = string(updatedname, "_")
+    end
+    push!(newnames, updatedname)
+  end
```
juliohm, line 52:
```suggestion
  for name in names
    n = name
    while n ∈ newnames
      n = string(n, "_")
    end
    push!(newnames, n)
  end
```

src/transforms/stdnames.jl
```diff
@@ -0,0 +1,61 @@
...
+
+  newnames
+end
+
+_camel(name) = join(uppercasefirst.(split(strip(name, delim), delim)))
+
+_snake(name) = lowercase(join(split(strip(name, delim), delim), '_'))
```
juliohm, line 59: Can you follow the same pattern of the _camel function? That is, use join at the end after the lowercase is applied to every word.

src/transforms/stdnames.jl
```diff
@@ -0,0 +1,61 @@
...
+
+_upper(name) = replace(uppercase(strip(name, delim)), delim => "")
```
juliohm, line 61: Do we really need the `strip` when we have `replace`?

test/transforms.jl
```diff
@@ -694,6 +694,68 @@
     @test n1 == n2
   end
 
+  @testset "StdNames" begin
+    # camel test
```
juliohm, line 698: This is not the camel test, please update or remove the comment.

test/transforms.jl
```diff
@@ -694,6 +694,68 @@
...
+    end
+
+    names = ["a", "A", "_a", "_A", "a ", "A "]
+    for name in names
+      @test TableTransforms._camel(name) == "A"
+      @test TableTransforms._snake(name) == "a"
+      @test TableTransforms._upper(name) == "A"
+    end
+
+    # invariant test
```
juliohm, line 713:
```suggestion
    # invariance test
```

juliohm (changes requested):

src/transforms/stdnames.jl
```diff
@@ -0,0 +1,61 @@
+# ------------------------------------------------------------------
+# Licensed under the MIT License. See LICENSE in the project root.
+# ------------------------------------------------------------------
+
+"""
+    StdNames(:spec)
+
+StdNames transform standardizes the column names.
```
juliohm, line 8: Please improve the docstring and write a text that helps end users understand the options. Try to follow the format of other docstrings.

juliohm:
@ceferisbarov please add the transform to the docs as usual.

src/transforms/stdnames.jl
```diff
@@ -0,0 +1,69 @@
...
+    while n ∈ newnames
+      n = string(n, "_")
+    end
+    push!(newnames, n)
+  end
+
+  newnames
+end
+
+_camel(name) = join(uppercasefirst.(split(strip(name, delim), delim)))
```
juliohm, line 65: Do we really the strip here?

juliohm (changes requested):
Can you also add tests for columns with special characters such as `A-B`, `A#`, etc?

src/transforms/stdnames.jl
ceferisbarov: We need it for snake case, otherwise, for example, "_A" gets split into "" and "A" and when we combine them, we get "_A" again.
  
Not sure why added this for camel case, TBH.

src/transforms/stdnames.jl
```diff
@@ -0,0 +1,69 @@
...
+    StdNames(:spec)
+
+The transform that standardizes the column names. There are three different methods of standardization: uppercase, camelcase, and snakecase.
+They can be specified by the `spec`` argument (`:upper`, `:camel`, and `:snake`, respectively). Transform defaults to uppercase. 
+
+# Examples
+
+```julia
+StdNames()
+StdNames(:camel)
```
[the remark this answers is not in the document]
ceferisbarov: I tried something with tables.

src/transforms/stdnames.jl
```diff
@@ -0,0 +1,73 @@
...
+
+# Examples
+
+| Transform | Original Col. Name | Transformed Col. Name |
+| ----------- | ----------- |	----------- |
+| `StdNames`    		 | apple_Trees | APPLETREES |
+| `StdNames(:upper)` | apple_Trees | APPLETREES |
+| `StdNames(:camel)` | apple_Trees | AppleTrees |
+| `StdNames(:snake)` | apple_Trees | apple_trees |
+
```
juliohm, line 20:
```suggestion
    StdNames(spec)

Standardizes column names according to given `spec`.
Default to `:upper` case specification.

# Specs

* `:upper` - Uppercase, e.g. COLUMNNAME
* `:camel` - Camelcase, e.g. ColumnName
* `:snake` - Snakecase, e.g. column_name
```

src/transforms/stdnames.jl
```diff
@@ -0,0 +1,73 @@
...
+  newtable, (rtrans, rcache)
+end
+
+function revert(::StdNames, newtable, cache)
+  rtrans, rcache = cache
+  revert(rtrans, newtable, rcache)
+end
+
+const delim = ['_', ' ']
```
juliohm, line 54: Can you please move this variable to near where it is used? Right before the _camel function. Please try to think like the reader of the code always.

juliohm: Also, we should probably add `-` to the list of delimiters. So that `day-of-month` gets converted into `DayOfMonth`

test/transforms.jl
```diff
@@ -694,6 +694,84 @@
...
+
+    name = "apple-tree"
+    @test TableTransforms._camel(name) == "Apple-tree"
+    @test TableTransforms._snake(name) == "apple-tree"
+    @test TableTransforms._upper(name) == "APPLE-TREE"
+    
+    name = "apple#"
+    @test TableTransforms._camel(name) == "Apple#"
+    @test TableTransforms._snake(name) == "apple#"
+    @test TableTransforms._upper(name) == "APPLE#"
```
juliohm, line 727: I think we should erase all special characters (except the `delim` characters) in a pre-processing step inside the `apply` function. @eliascarv do you also think this is a good idea?

I think we should set `delim = [' ', '\t', '-', '_']` as our possible delimiters. All other characters could be erased.

ceferisbarov: I would also suggest adding `&` to the list of delimiters. That being said, I don't think we can make this transform work well with complex column names, anyway. Maybe we can later add more specific transforms that deal with column names (such as `Replace`, but for column names). @juliohm @eliascarv

juliohm:
As I mentioned it should be easy to erase all special symbols as a pre
processing step with a Regex for example. We just need to decide the valids
delimiters. I don't think & is widely used in statistics as a delimiter. It
is used as an interaction term.

Em qua., 11 de mai. de 2022 05:46, Jafar Isbarov ***@***.***>
escreveu:

src/transforms/stdnames.jl
```diff
@@ -0,0 +1,76 @@
...
+* `:snake` - Snakecase, e.g. column_name
+"""
+struct StdNames <: Stateless
+  spec::Symbol
+end
+
+StdNames() = StdNames(:upper)
+
+isrevertible(::Type{StdNames}) = true
+
```
juliohm, line 24:
```suggestion
```

src/transforms/stdnames.jl
```diff
@@ -0,0 +1,76 @@
...
+function revert(::StdNames, newtable, cache)
+  rtrans, rcache = cache
+  revert(rtrans, newtable, rcache)
+end
+
+function _filter(name)
+  _keep(char) = isdigit(char) || isletter(char) || occursin(char, String(delim))
+  filter(char -> _keep(char), name)
+end
+
```
juliohm, line 55:
```suggestion
_clean(name) = filter(c -> isdigit(c) || isletter(c) || c ∈ delim, name)
```

src/transforms/stdnames.jl
```diff
@@ -0,0 +1,76 @@
...
+function apply(transform::StdNames, table)  
+  cols = Tables.columns(table)
+  oldnames = string.(Tables.columnnames(cols))
+  preprocessed = _filter.(oldnames)
+  
+  spec = transform.spec
+
+  spec == :camel && (names = _camel.(preprocessed))
+  spec == :snake && (names = _snake.(preprocessed))
+  spec == :upper && (names = _upper.(preprocessed))
```
juliohm, line 35:
```suggestion
  cleaned = _clean.(oldnames)
  
  spec = transform.spec

  spec == :camel && (names = _camel.(cleaned))
  spec == :snake && (names = _snake.(cleaned))
  spec == :upper && (names = _upper.(cleaned))
```

juliohm: Always try to choose short names for variables.

src/transforms/stdnames.jl
```diff
@@ -0,0 +1,76 @@
...
+    while n ∈ newnames
+      n = string(n, "_")
+    end
+    push!(newnames, n)
+  end
+
+  newnames
+end
+
+const delim = [' ', '\t', '-', '_']
```
juliohm, line 70: Now that this delim variable is used in _clean, it needs to appear earlier in the file. Someone reading the code will try to find where it is defined. Think in terms of readers always.

juliohm (changes requested):
I think we are almost done with this one @ceferisbarov , can you please take a look at this new set of revisions?

src/transforms/stdnames.jl
```diff
@@ -0,0 +1,71 @@
...
+function apply(transform::StdNames, table)  
+  cols = Tables.columns(table)
+  oldnames = string.(Tables.columnnames(cols))
+  cleaned = _clean.(oldnames)
+  
+  spec = transform.spec
+
+  spec == :camel && (names = _camel.(cleaned))
+  spec == :snake && (names = _snake.(cleaned))
+  spec == :upper && (names = _upper.(cleaned))
```
juliohm, line 34:
```suggestion
  # retrieve spec
  spec = transform.spec
  
  # retrieve column names
  cols = Tables.columns(table)
  oldnames = string.(Tables.columnnames(cols))
  
  # clean column names
  cleaned = _clean.(oldnames)

  # apply spec
  spec == :camel && (names = _camel.(cleaned))
  spec == :snake && (names = _snake.(cleaned))
  spec == :upper && (names = _upper.(cleaned))
```

src/transforms/stdnames.jl
```diff
@@ -0,0 +1,71 @@
...
+
+  newnames = _unique(names)
+  oldnew = Dict(zip(oldnames, newnames))
+
+  rtrans = Rename(oldnew)
+  newtable, rcache = apply(rtrans, table)
```
juliohm, line 40:
```suggestion
  # make names unique
  newnames = _unique(names)
  
  # rename transform
  rtrans = Rename(Dict(oldnames .=> newnames))
  newtable, rcache = apply(rtrans, table)
```

juliohm (changes requested):

juliohm:
@eliascarv anything else I missed before we can merge this?

eliascarv (approved):

eliascarv:
For me everything is excellent.

juliohm:
Another great addition @ceferisbarov ! 🎉

Issue #76, opened 2022-05-17
Reverting `Select` inside pipeline fails if columntypes differ


antholzer:
For example in the following
```julia
df = (; a=rand(Float32, 2), b=rand(2))
p = Select(:b) → MinMax()
newtable, cache = apply(p, df)
revert(p, newtable, cache)
```
the column a gets converted into a `Float64` vector
If one uses the following table `df = (; a=[rand(3), rand(3)], b=rand(2))` the inversion fails since since it tries to convert `Vector{Float64}` to `Float64`.

Is this behavior intentional? 

Explicitly creating and `Any` vector should fix it i.e.
```
diff --git a/src/transforms/select.jl b/src/transforms/select.jl
index 5748372..09ae813 100644
--- a/src/transforms/select.jl
+++ b/src/transforms/select.jl
@@ -108,7 +108,7 @@ function revert(::Select, newtable, cache)
   # selected columns
   cols   = Tables.columns(newtable)
   select = Tables.columnnames(cols)
-  scols  = [Tables.getcolumn(cols, name) for name in select]
+  scols  = Any[Tables.getcolumn(cols, name) for name in select]
 
   # rejected columns
   reject, rcols, sperm, rinds = cache
```

Shall I open a pull request for this fix?

juliohm:
Thank you for reporting the bug @antholzer , we should definitely fix it. I wonder if adding `Any` is the right solution though. I don't have time to look into the code right now, but do you mean that the result of Select would always lead to `Any` columns in that case? Do we have an alternative solution?

eliascarv:
The error is not in `Select` but in `Scale`. It is providing the column type to Float64 in its operations.

juliohm:
Good catch @eliascarv , would you like to propose a fix? I am super busy here...

antholzer:
Actually bot have an error with type conversion.
```
df = (; a=rand(Float32, 2), b=rand(2))
p=Select(:b) → DropMissing()
newtable, cache = apply(p, df)
revert(p, newtable, cache)
```
output:
```
# master
(a = [0.5341609716415405, 0.6113829612731934], b = [0.6949850019609314, 0.4067167096075357])
# with `Any`
(a = Float32[0.534161, 0.61138296], b = [0.6949850019609314, 0.4067167096075357])
```

`Any` is for the vector of all the columns i.e. the individual columns are still their respective types.

[...]

PR #77, opened 2022-05-18
fix: scale no longer converts all floats to Float64

antholzer:
As noted by @eliascarv in https://github.com/JuliaML/TableTransforms.jl/issues/76#issuecomment-1128835713 `Scale` always promotes column types to `Float64`. This should fix this issue and make it work with different float types.

### DIFF
--- a/src/transforms/scale.jl
+++ b/src/transforms/scale.jl
@@ -21,15 +21,16 @@ Scale(low=0.3, high=0.7)
 
 * The `low` and `high` values are restricted to the interval [0, 1].
 """
-struct Scale{T} <: Colwise
+struct Scale{T<:Real} <: Colwise
   low::T
   high::T
 
-  function Scale(low::T1, high::T2) where {T1,T2}
+  function Scale(low::T, high::T) where {T<:Real}
     @assert 0 ≤ low ≤ high ≤ 1 "invalid quantiles"
-    new{promote_type(T1, T2)}(low, high)
+    new{T}(low, high)
   end
 end
+Scale(low::Real, high::Real) = Scale(promote(low, high)...)
 
 Scale(; low=0.25, high=0.75) = Scale(low, high)
 
--- a/src/transforms/scale.jl
+++ b/src/transforms/scale.jl
@@ -30,6 +30,7 @@ struct Scale{T<:Real} <: Colwise
     new{T}(low, high)
   end
 end
+
 Scale(low::Real, high::Real) = Scale(promote(low, high)...)
 
 Scale(; low=0.25, high=0.75) = Scale(low, high)
--- a/src/transforms/scale.jl
+++ b/src/transforms/scale.jl
@@ -21,21 +21,21 @@ Scale(low=0.3, high=0.7)
 
 * The `low` and `high` values are restricted to the interval [0, 1].
 """
-struct Scale <: Colwise
-  low::Float64
-  high::Float64
+struct Scale{T} <: Colwise
+  low::T
+  high::T
 
-  function Scale(low, high)
+  function Scale(low::T1, high::T2) where {T1,T2}
     @assert 0 ≤ low ≤ high ≤ 1 "invalid quantiles"
-    new(low, high)
+    new{promote_type(T1, T2)}(low, high)
   end
 end
 
 Scale(; low=0.25, high=0.75) = Scale(low, high)
 
-assertions(::Type{Scale}) = [assert_continuous]
+assertions(::Type{<:Scale}) = [assert_continuous]
 
-isrevertible(::Type{Scale}) = true
+isrevertible(::Type{<:Scale}) = true
 
 function colcache(transform::Scale, x)
   levels = (transform.low, transform.high)
@@ -55,7 +55,7 @@ The transform that is equivalent to `Scale(low=0, high=1)`.
 
 See also [`Scale`](@ref).
 """
-MinMax() = Scale(low=0.0, high=1.0)
+MinMax() = Scale(low=0, high=1)
 
 """
     Interquartile()
--- a/test/transforms.jl
+++ b/test/transforms.jl
@@ -1072,6 +1072,17 @@
     @test Tables.isrowtable(n)
     rtₒ = revert(T, n, c)
     @test Tables.matrix(rt) ≈ Tables.matrix(rtₒ)
+
+    # columntype does not change
+    for FT in (Float16, Float32)
+      t = Table(; x=rand(FT, 10))
+      for T in (MinMax(), Scale(FT(0), FT(0.5)))
+        n, c = apply(T, t)
+        @test Tables.columntype(t, :x) == Tables.columntype(n, :x)
+        tₒ = revert(T, n, c)
+        @test Tables.columntype(t, :x) == Tables.columntype(tₒ, :x)
+      end
+    end
   end
 
   @testset "ZScore" begin


test/transforms.jl
```diff
@@ -951,12 +951,14 @@
 
   @testset "Scale" begin
     # constant column
-    x = fill(3.0, 10)
+    x = fill(Float64(3.0), 10)
```
juliohm, line 954: Please erase this change as it doesn't change the results.

test/transforms.jl
```diff
@@ -951,12 +951,14 @@
...
     y = rand(10)
-    t = Table(; x, y)
+    t = Table(; x=x, y=y, xf32=Float32.(x), xf16=Float16.(x))
```
juliohm, line 956: We don't alter existing tests in PRs. Please write a separate test for the fix you are proposing.

juliohm (changes requested):
Left a first round of reviews @antholzer

test/transforms.jl
juliohm: `3.0` is always `Float64` in Julia. For single precision you write `3.0f0`. Integers behave differently in computers: if you specify the type `Int` in your code, then it will be dependent on the word size, either `Int32` or `Int64`.

test/transforms.jl
antholzer: I added a separate test to check that the eltype of the column does not change, but still in the same testset. Is that what you wanted?

src/transforms/scale.jl
```diff
@@ -38,7 +38,7 @@ assertions(::Type{Scale}) = [assert_continuous]
 isrevertible(::Type{Scale}) = true
 
 function colcache(transform::Scale, x)
-  levels = (transform.low, transform.high)
+  levels = map(eltype(x), (transform.low, transform.high))
```
juliohm, line 41: I started thinking that maybe we shouldn't change the code here, but instead ask users to create a Scale transform with the appropriate type. `Scale(0.0f0, 1.0f0)` if they want to preserve the result type.

@eliascarv what do you think? I feel that converting the type explicitly like in the proposed change can lead to more issues when the eltype of the column is not trivial. For example, if the eltype is Integer, we will loose precision, etc.

juliohm: So I think a better fix would consist of adding a type parameter to Scale{T}.

eliascarv: I think this problem deserves a more general solution for all Colwise.

juliohm: That is a good point as well. I think we can still improve the situation with the Scale transform. It is too rigid at the moment requireing Float64 fields. @antholzer can you parametrize the transform to accept low::T and high::T?

antholzer: I added a type parameter `T`.

For `MinMax` I used `low=0, high=1` instead of `low=0.0, high=1.0` since we do not lose precision and this way all Vectors with `Float16`, `Float32`, `Float64` or `Int`entries do not get converted.

juliohm: Awesome @antholzer , thank you for the great improvements. Will leave another review now.

src/transforms/scale.jl
```diff
@@ -21,32 +21,32 @@ Scale(low=0.3, high=0.7)
...
     @assert 0 ≤ low ≤ high ≤ 1 "invalid quantiles"
-    new(low, high)
+    new{promote_type(T1, T2)}(low, high)
   end
 end
 
 Scale(; low=0.25, high=0.75) = Scale(low, high)
 
-assertions(::Type{Scale}) = [assert_continuous]
+assertions(::Type{Scale{T}}) where {T} = [assert_continuous]
```
juliohm, line 36:
```suggestion
assertions(::Type{<:Scale}) = [assert_continuous]
```

src/transforms/scale.jl
```diff
@@ -21,32 +21,32 @@ Scale(low=0.3, high=0.7)
...
 
-isrevertible(::Type{Scale}) = true
+isrevertible(::Type{Scale{T}}) where {T} = true
```
juliohm, line 38:
```suggestion
isrevertible(::Type{<:Scale}) = true
```

src/transforms/scale.jl
```diff
@@ -21,32 +21,32 @@ Scale(low=0.3, high=0.7)
...
-function colcache(transform::Scale, x)
+function colcache(transform::Scale{T}, x) where {T}
   levels = (transform.low, transform.high)
   xl, xh = quantile(x, levels)
   xl == xh && ((xl, xh) = (zero(xl), one(xh)))
   (xl=xl, xh=xh)
 end
 
-colapply(::Scale, x, c)  = @. (x - c.xl) / (c.xh - c.xl)
+colapply(::Scale{T}, x, c) where{T}  = @. (x - c.xl) / (c.xh - c.xl)
```
juliohm, line 47:
```suggestion
colapply(::Scale, x, c)  = @. (x - c.xl) / (c.xh - c.xl)
```

src/transforms/scale.jl
```diff
@@ -21,32 +21,32 @@ Scale(low=0.3, high=0.7)
...
 
-colrevert(::Scale, y, c) = @. (c.xh - c.xl) * y + c.xl
+colrevert(::Scale{T}, y, c) where{T} = @. (c.xh - c.xl) * y + c.xl
```
juliohm, line 49:
```suggestion
colrevert(::Scale, y, c) = @. (c.xh - c.xl) * y + c.xl
```

src/transforms/scale.jl
```diff
@@ -21,21 +21,21 @@ Scale(low=0.3, high=0.7)
...
+  low::T
+  high::T
 
-  function Scale(low, high)
+  function Scale(low::T1, high::T2) where {T1,T2}
     @assert 0 ≤ low ≤ high ≤ 1 "invalid quantiles"
-    new(low, high)
+    new{promote_type(T1, T2)}(low, high)
   end
 end
```
eliascarv, line 32: This is a safer design for this struct:
```julia
struct Scale{T<:AbstractFloat} <: Colwise
  low::T
  high::T
end

Scale(low::AbstractFloat, high::AbstractFloat) = 
  Scale(promote(low, high)...)
  
Scale(low::Real, high::Real) = Scale(float(low), float(high))
```

eliascarv:
The low and high values ​​are restricted to the interval [0, 1]. However, we can type them as Reals or nNumbers instead of Floats.

juliohm:
So the PR used Int for the MinMax transform. You are suggesting that these
Int should be promoted to float? I'm far from my computer to read this
carefully...

Em qua., 18 de mai. de 2022 13:35, Elias Carvalho ***@***.***>
escreveu:

eliascarv:
No, I just used `AbstractFloat` because of the original code, but the `quantile` function accepts real numbers.

eliascarv:
> The low and high values ​​are restricted to the interval [0, 1]. However, we can type them as `Real` or `Number` instead of `AbstractFloat`:
> 
> ```julia
> struct Scale{T<:Real} <: Colwise
>   low::T
>   high::T
>   function Scale(low::T, high::T) where {T<:Real}
>     @assert 0 ≤ low ≤ high ≤ 1 "Invalid quantiles."
>     new{T}(low, high)
>   end
> end
> 
> Scale(low::Real, high::Real) = Scale(promote(low, high)...)
> ```
> 
> What do you think @juliohm?

This last version is more generic.

juliohm:
@eliascarv the proposed change consists of restricting the low and high percentiles to Real right?  I understand that the intention is to throw an error whenever someone tries to construct percentiles that are not Real. Makes sense to me.

@antholzer can you follow the suggestion and restrict T<:Real in the PR?

eliascarv:
> The low and high values ​​are restricted to the interval [0, 1]. However, we can type them as `Real` or `Number` instead of `AbstractFloat`:
> 
> ```julia
> struct Scale{T<:Real} <: Colwise
>   low::T
>   high::T
>   function Scale(low::T, high::T) where {T<:Real}
>     @assert 0 ≤ low ≤ high ≤ 1 "Invalid quantiles."
>     new{T}(low, high)
>   end
> end
> 
> Scale(low::Real, high::Real) = Scale(promote(low, high)...)
> ```
> 
> What do you think @juliohm?

@antholzer follow this 👆

src/transforms/scale.jl
```diff
@@ -21,21 +21,22 @@ Scale(low=0.3, high=0.7)
...
+  high::T
 
-  function Scale(low, high)
+  function Scale(low::T, high::T) where {T<:Real}
     @assert 0 ≤ low ≤ high ≤ 1 "invalid quantiles"
-    new(low, high)
+    new{T}(low, high)
   end
 end
+Scale(low::Real, high::Real) = Scale(promote(low, high)...)
```
juliohm, line 33:
```suggestion

Scale(low::Real, high::Real) = Scale(promote(low, high)...)
```

juliohm:
Thank you @antholzer for another awesome improvement. I will trigger a minor release now with the bug fixes and with the new StdNames transform.

PR #82, opened 2022-05-23
Add `ndim` option to `EigenAnalysis`

### LINKED ISSUE #17: Add dim option to EigenAnalysis to reduce the number of dimensions
Currently we retain all scores. It would be nice to have an option to retain only first `dim` elements in the basis.

eliascarv:
This PR adds the `ndim` keyword argument to `EigenAnalysis`, it represents the number of dimensions of the output.

closes #17


eliascarv:
Note: I removed "center the data" because `ZScore` does that.

src/transforms/eigenanalysis.jl
```diff
@@ -2,10 +2,13 @@
 # Licensed under the MIT License. See LICENSE in the project root.
 # ------------------------------------------------------------------
 
+const NDim = Union{Int,Colon}
+
```
juliohm, line 6: Please avoid creating these isolated constructs. Simply pass the union type directly to the EigenAnalysis struct or simply don't constrain the type T. EigenAnalysis{T} <:  Transform should be enough.

src/transforms/eigenanalysis.jl
```diff
@@ -31,21 +34,30 @@ for more details about these three variants of eigenanalysis.
...
 EigenAnalysis(:VDV)
+EigenAnalysis(:V, ndim=2)
+EigenAnalysis(:VD, ndim=3)
+EigenAnalysis(:VDV, ndim=4)
 ```
 """
-struct EigenAnalysis <: Transform
+struct EigenAnalysis{T<:NDim} <: Transform
   proj::Symbol
+  ndim::T
```
juliohm, line 44:
```suggestion
struct EigenAnalysis <: Transform
  proj::Symbol
  ndim::Union{Int,Nothing}
```

src/transforms/eigenanalysis.jl
```diff
@@ -54,47 +66,46 @@ function apply(transform::EigenAnalysis, table)
...
   cols = Tables.columns(table)
-  names = Tables.columnnames(cols)
+  onames = Tables.columnnames(cols)
 
   # table as matrix
   X = Tables.matrix(table)
 
-  # center the data
-  μ = mean(X, dims=1)
-  Y = X .- μ
```
juliohm, line 64: This operation is necessary inside EigenAnalysis. The other transforms PCA and SDS etc do additional steps, but this is irrelevant here.

juliohm (changes requested):
Try to reduce this PR to a minimum amount of changes. You are changing current behavior by removing the centering operation and introducing a colon option. Keep it simple: an optional argument ndim that defaults to nothing. The default is the current behavior and if the user passes an integer, just select the columns.

src/transforms/eigenanalysis.jl
```diff
@@ -31,21 +34,30 @@ for more details about these three variants of eigenanalysis.
...
-  function EigenAnalysis(proj)
-    @assert proj ∈ (:V, :VD, :VDV) "invalid projection"
-    new(proj)
+  function EigenAnalysis(proj::Symbol, ndim::T) where {T<:NDim}
+    @assert proj ∈ (:V, :VD, :VDV) "Invalid projection."
+    new{T}(proj, ndim)
   end
 end
 
+EigenAnalysis(proj::Symbol; ndim::NDim=:) = EigenAnalysis(proj, ndim)
```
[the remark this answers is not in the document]
eliascarv: In Julia, the colon operator means to select all elements, that's why I used it.
But, I understand your points, so I'm going to replace it.

eliascarv:
@juliohm, I had to update the benchmarks probably because the eigenvalues and eigenvectors order has changed.

src/transforms/eigenanalysis.jl
```diff
@@ -3,9 +3,10 @@
 # ------------------------------------------------------------------
 
 """
-    EigenAnalysis(proj)
+    EigenAnalysis(proj; ndim=nothing)
```
juliohm, line 6: Let's make it a positional argument. This EigenAnalysis transform is more low-level and someone using it can simply pass a second argument if necessary without typing ndim.

src/transforms/eigenanalysis.jl
```diff
@@ -31,21 +32,26 @@ for more details about these three variants of eigenanalysis.
 EigenAnalysis(:V)
 EigenAnalysis(:VD)
 EigenAnalysis(:VDV)
+EigenAnalysis(:V, ndim=2)
```
juliohm, line 35:
```suggestion
EigenAnalysis(:V, 2)
```

src/transforms/eigenanalysis.jl
```diff
@@ -31,21 +32,26 @@ for more details about these three variants of eigenanalysis.
...
 ```
 """
 struct EigenAnalysis <: Transform
   proj::Symbol
+  ndim::Union{Int,Nothing}
 
-  function EigenAnalysis(proj)
-    @assert proj ∈ (:V, :VD, :VDV) "invalid projection"
-    new(proj)
+  function EigenAnalysis(proj; ndim=nothing)
```
juliohm, line 42:
```suggestion
  function EigenAnalysis(proj, ndim=nothing)
```

src/transforms/eigenanalysis.jl
```diff
@@ -69,32 +78,35 @@ function apply(transform::EigenAnalysis, table)
   # project the data
   Z = Y * S
 
+  # discarted and selected coluns
```
juliohm, line 81:
```suggestion
  # discard columns
```

src/transforms/eigenanalysis.jl
```diff
@@ -69,32 +78,35 @@ function apply(transform::EigenAnalysis, table)
...
+  D = Z[:, ndim+1:end]
```
juliohm, line 82: I think we don't need to introduce a new matrix in the cache. We can simply post-process the matrix S and S^-1 to discard columns right?

src/transforms/eigenanalysis.jl
```diff
@@ -148,31 +167,52 @@ function eigenmatrices(transform, Y)
 end
 
 """
-    PCA()
+    PCA(; ndim=nothing)
```
juliohm, line 170: Let's make it a positional argument. It doesn't have any argument so there is no confusion.

test/transforms.jl
```diff
@@ -1333,12 +1333,12 @@
...
-      p₄ = scatter(t₄.x, t₄.y, label="VDV")
-      p₅ = scatter(t₅.x, t₅.y, label="PCA")
-      p₆ = scatter(t₆.x, t₆.y, label="DRS")
-      p₇ = scatter(t₇.x, t₇.y, label="SDS")
+      p₂ = scatter(t₂.pc2, t₂.pc1, label="V")
+      p₃ = scatter(t₃.pc2, t₃.pc1, label="VD")
+      p₄ = scatter(t₄.pc2, t₄.pc1, label="VDV")
+      p₅ = scatter(t₅.pc2, t₅.pc1, label="PCA")
+      p₆ = scatter(t₆.pc2, t₆.pc1, label="DRS")
+      p₇ = scatter(t₇.pc2, t₇.pc1, label="SDS")
```
juliohm, line 1341: Let's make the variable names upper case: PC1, PC2, ...

juliohm (changes requested):
Please try to modify the transformation matrices S and S inverse instead. They fully define the projections and we just need to discard some columns.

Also, please make sure that the previous images in our visual tests aren't altered. This new feature should be backward compatible.

src/transforms/eigenanalysis.jl
eliascarv: You're totally right. I will change it.

eliascarv:
This PR is getting really messed up. I will open another one to continue the work.

PR #83, opened 2022-05-24
add Levels transform

Frank-III:
The way I currently store caches is to mix functions and vector which is absolutely not optimal, but I am not sure if I should only allow change the levels or order of `CategoricalArray` or should generalized for `Vector` as well. (I guess to allow changing `levels` is fine as we specified the levels to guarantee it is not an infinite column), and for orders when it is set to be true for a float column the `categorical` function still work but make no sense.


src/TableTransforms.jl
```diff
@@ -11,7 +11,8 @@ using Transducers: tcollect
 using LinearAlgebra
 using Statistics
 using PrettyTables
-
+using CategoricalArrays
+import Base.@kwdef
```
juliohm, line 15: Please add `using CategoricalArrays` in the block containing the other using statements, and drop the `import Base.@kwdef`. It is easy enough to write the struct outer constructor with keyword args manually, and we avoid adding a dependency on this macro.

src/TableTransforms.jl
```diff
@@ -49,6 +50,6 @@ export
   PCA, DRS, SDS,
   Sequential,
   Parallel,
-  →, ⊔
-
+  →, ⊔,
+  Levels
```
juliohm, line 54: Please add this transform to another place in the list. The list of transforms that is exported has an implicit logic to it. Sequential and Parallel should be the last transforms.

src/transforms.jl
```diff
@@ -228,4 +228,5 @@ include("transforms/quantile.jl")
 include("transforms/functional.jl")
 include("transforms/eigenanalysis.jl")
 include("transforms/sequential.jl")
-include("transforms/parallel.jl")
\ No newline at end of file
+include("transforms/parallel.jl")
+include("transforms/levels.jl")
```
juliohm, line 232: Same here, please include the file in a different place in the list. The sequential and parallel should be the last.

src/transforms/levels.jl
```diff
@@ -0,0 +1,72 @@
+"""
+    Levels(:a => ["yes,"no"])
+
+Return a copy of the table with specified levels and orders for categorical columns
+allowing only changing the order of the column
```
juliohm, line 5:
```suggestion
allowing only changing the order of the column.
```

src/transforms/levels.jl
```diff
@@ -0,0 +1,72 @@
...
+
+# Examples
+
+```julia
+Levels(:a => ["yes,"no"], :c => [1,2,4], :d = ["a","b","c"])
```
juliohm, line 10:
```suggestion
Levels(:a => ["yes,"no"], :c => [1,2,4], :d => ["a","b","c"])
```

src/transforms/levels.jl
```diff
@@ -0,0 +1,72 @@
...
+Levels("a" => ["yes","no"], "c" => [1,2,4], ordered= ["a","c"])
```
juliohm, line 11:
```suggestion
Levels("a" => ["yes","no"], "c" => [1,2,4], ordered = ["a","c"])
```

src/transforms/levels.jl
```diff
@@ -0,0 +1,72 @@
...
+Levels(:a => ["yes","no"], :c => [1,2,4], :d => [1,23,5,7] ,ordered = [:a,:b,:c])
```
juliohm, line 12:
```suggestion
Levels(:a => ["yes","no"], :c => [1,2,4], :d => [1,23,5,7] , ordered = [:a,:b,:c])
```
Please be more careful with spaces, they matter.

src/transforms/levels.jl
```diff
@@ -0,0 +1,72 @@
...
+```
+"""
+
```
juliohm, line 15:
```suggestion
```

src/transforms/levels.jl
```diff
@@ -0,0 +1,72 @@
...
+@kwdef struct Levels{K} <: Stateless
+  levelspec::K
+  ordered::AbstractVector{Symbol}
+end
```
juliohm, line 19: Please remove the @kwdef dependency and use another type for the ordered field that is a concrete type, for example Vector{Symbol}

src/transforms/levels.jl
```diff
@@ -0,0 +1,72 @@
...
+#should work more on constructors
+Levels(pairs::Pair{K}...; ordered::AbstractVector{K}=Symbol[]) where {K <: Symbol} =
+  Levels(NamedTuple(pairs), ordered)
+
+Levels(pairs::Pair{K}...; ordered::AbstractVector{K}=String[]) where {K <: AbstractString} =
+  Levels(NamedTuple(Symbol(k) => v for (k,v) in pairs), Symbol.(ordered))
+
+
+
```
juliohm, line 28:
```suggestion
```
Whitespace matters.

src/transforms/levels.jl
```diff
@@ -0,0 +1,72 @@
...
+isrevertible(transform::Levels) = true
```
juliohm, line 29:
```suggestion
isrevertible(transform::Levels) = true

```

src/transforms/levels.jl
```diff
@@ -0,0 +1,72 @@
...
+# handle three cases
+#1. when the col is already a categorical array and wanna change levels and order
+categorify(level::AbstractVector, x::CategoricalVector, ordered::Bool) = levels(x), categorical(x, levels = level, ordered=ordered) 
+#2. when the col is normal array and want to change to categorical array
+categorify(level::AbstractVector, x::AbstractVector, ordered::Bool) = unwrap, categorical(x,levels = level, ordered=ordered) 
+#3. when the col is not need for change or convert back to normal array
+categorify(func::Function, x::AbstractVector, ordered::Bool) = ordered ? (levels(x), categorical(x, ordered=true)) : (func, func.(x))
+
```
juliohm, line 37:
```suggestion
```

src/transforms/levels.jl
```diff
@@ -0,0 +1,72 @@
...
+
+isrevertible(transform::Levels) = true
+# handle three cases
+#1. when the col is already a categorical array and wanna change levels and order
+categorify(level::AbstractVector, x::CategoricalVector, ordered::Bool) = levels(x), categorical(x, levels = level, ordered=ordered) 
+#2. when the col is normal array and want to change to categorical array
+categorify(level::AbstractVector, x::AbstractVector, ordered::Bool) = unwrap, categorical(x,levels = level, ordered=ordered) 
+#3. when the col is not need for change or convert back to normal array
+categorify(func::Function, x::AbstractVector, ordered::Bool) = ordered ? (levels(x), categorical(x, ordered=true)) : (func, func.(x))
```
juliohm, line 36:
```suggestion
# when the col is already a categorical array and wanna change levels and order
categorify(l, x::CategoricalVector, o) = levels(x), categorical(x, levels=l, ordered=o)

# when the col is normal array and want to change to categorical array
categorify(l, x::AbstractVector, o) = unwrap, categorical(x, levels=l, ordered=o)

# when the col is not need for change or convert back to normal array
categorify(f::Function, x::AbstractVector, o) = o ? (levels(x), categorical(x, ordered=true)) : (f, f.(x))
```

src/transforms/levels.jl
```diff
@@ -0,0 +1,72 @@
...
+  new_levels = transform.levelspec
+  caches = Vector{Union{Vector,Function}}()
+  ncols = map(names) do nm
+    x = Tables.getcolumn(cols, nm)
+    level_ = get(new_levels, nm, identity)
+    ordered = in(nm,transform.ordered)
+    cache, new_x = categorify(level_, x, ordered)
+    push!(caches, cache)
+    new_x
+  end
```
juliohm, line 51: We don't use underscores in our variable names.

src/transforms/levels.jl
```diff
@@ -0,0 +1,72 @@
...
+  
+  𝒯 = (; zip(names, ncols)...)
+  newtable = 𝒯 |> Tables.materializer(table)
+  
+  return newtable, caches
```
juliohm, line 56:
```suggestion
  newtable, caches
```

test/transforms.jl
```diff
@@ -1,6 +1,45 @@
 @testset "Transforms" begin
   # using MersenneTwister for compatibility between Julia versions
   rng = MersenneTwister(42)
+  @testset "Levels" begin
+    a = categorical(["yes","no","no","no","yes"]) 
+    b = categorical([1,2,4,2,8],ordered=false) 
+    c =categorical([1,2,1,2,1]) 
+    d= [1,23,5,7,7]
```
juliohm, line 8:
```suggestion
    d = [1,23,5,7,7]
```

test/transforms.jl
```diff
@@ -1,6 +1,45 @@
...
+    t = Table(; a, b, c, d)
+
+    T = Levels(:a => ["yes","no"], :c => [1,2,4], :d => [1,23,5,7])
+    T_same = Levels("a" => ["yes","no"], "c" => [1,2,4], "d" => [1,23,5,7])
```
juliohm, line 12: Please remove T_same as it is the same test, except the constructor is different. If you want to test the constructor, please use a separate block of tests.

test/transforms.jl
```diff
@@ -1,6 +1,45 @@
...
+    n,c = apply(T,t)
+    n,c = apply(T_same,t)
+    @test levels(n.a) == ["yes","no"]
+    @test levels(n.c) == [1,2,4]
+    tₒ = revert(T, n, c)
+    @test levels(tₒ.a) == ["no","yes"]
+    @test levels(tₒ.c) == [1,2]
+
+    T = Levels(:a => ["yes","no"], :c => [1,2,4], :d => [1,23,5,7],ordered=[:b])
+    T_same = Levels("a" => ["yes","no"], "c" => [1,2,4], "d" => [1,23,5,7],ordered=["b"])
```
juliohm, line 22: Same here.

test/transforms.jl
```diff
@@ -1,6 +1,45 @@
...
+    T_same = Levels("a" => ["yes","no"], "c" => [1,2,4], "d" => [1,23,5,7])
+    n,c = apply(T,t)
+    n,c = apply(T_same,t)
+    @test levels(n.a) == ["yes","no"]
+    @test levels(n.c) == [1,2,4]
+    tₒ = revert(T, n, c)
+    @test levels(tₒ.a) == ["no","yes"]
+    @test levels(tₒ.c) == [1,2]
+
+    T = Levels(:a => ["yes","no"], :c => [1,2,4], :d => [1,23,5,7],ordered=[:b])
```
juliohm, line 21: Please review all whitespaces in this PR.
```suggestion
    T = Levels(:a => ["yes","no"], :c => [1,2,4], :d => [1,23,5,7], ordered=[:b])
```

juliohm (changes requested):
This is an awesome PR @Frank-III , thank you for the addition. Left a couple of review comments.

src/transforms/levels.jl
Frank-III: It seems that we have to specify the level type for the first one to be `AbstractVector` as when `identity` or `unwrap` function get called on `CategoricalArray` it would be ambiguous

juliohm: Thanks for checking. Please add the minimum number of type annotations that can disambiguate the methods.

Project.toml
```diff
@@ -4,6 +4,7 @@ authors = ["Júlio Hoffimann <julio.hoffimann@gmail.com> and contributors"]
 version = "1.1.0"
 
 [deps]
+CategoricalArrays = "324d7699-5711-5eae-9e2f-1d82baa6b597"
```
juliohm, line 7: I think we also need an entry in the [compat] section with the version of CategoricalArrays.jl introduced here. It should be the latest stable release of the package.

src/transforms/levels.jl
```diff
@@ -0,0 +1,70 @@
...
+Levels(:a => ["yes", "no"], :c => [1, 2, 4], :d => [1, 23, 5, 7], ordered = [:a, :b, :c])
+```
+"""
+struct Levels{K} <: Stateless
+  levelspec::K
+  ordered::Vector{Symbol}
+end
+
+Levels(pairs::Pair{K}...; ordered::AbstractVector{K}=Symbol[]) where {K <: Symbol} =
+  Levels(NamedTuple(pairs), ordered)
```
juliohm, line 21:
```suggestion
Levels(pairs::Pair{Symbol}...; ordered=Symbol[]) =
  Levels(NamedTuple(pairs), ordered)
```

src/transforms/levels.jl
```diff
@@ -0,0 +1,70 @@
...
+
+Levels(pairs::Pair{K}...; ordered::AbstractVector{K}=String[]) where {K <: AbstractString} =
+  Levels(NamedTuple(Symbol(k) => v for (k,v) in pairs), Symbol.(ordered))
```
juliohm, line 24:
```suggestion
Levels(pairs::Pair{K}...; ordered=K[]) where {K <: AbstractString} =
  Levels(NamedTuple(Symbol(k) => v for (k,v) in pairs), Symbol.(ordered))
```

src/transforms/levels.jl
```diff
@@ -0,0 +1,70 @@
...
+
+function apply(transform::Levels,table)
+  cols = Tables.columns(table)
+  names = Tables.columnnames(cols)
+  newlevels = transform.levelspec
+  caches = Vector{Union{Vector,Function}}()
+  ncols = map(names) do nm
+    x = Tables.getcolumn(cols, nm)
+    l = get(newlevels, nm, identity)
+    o = in(nm, transform.ordered)
```
juliohm, line 45: Please use \in as an infix operator.

src/transforms/levels.jl
```diff
@@ -0,0 +1,70 @@
...
+categorify(l, x::AbstractVector, o) = unwrap, categorical(x, levels=l, ordered=o) 
+
+# when the col is not need for change or convert back to normal array
+categorify(f::Function, x::AbstractVector, o) = o ? (levels(x), categorical(x, ordered=true)) : (f, f.(x))
+
+function apply(transform::Levels,table)
+  cols = Tables.columns(table)
+  names = Tables.columnnames(cols)
+  newlevels = transform.levelspec
+  caches = Vector{Union{Vector,Function}}()
```
juliohm, line 41: Can you avoid initialization with a specific type and exploit the do syntax below to create this variable instead?

src/transforms/levels.jl
```diff
@@ -0,0 +1,70 @@
...
+  ncols = map(names) do nm
+    x = Tables.getcolumn(cols, nm)
+    l = get(newlevels, nm, identity)
+    o = in(nm, transform.ordered)
+    cache, newx = categorify(l, x, o)
```
juliohm, line 46: The order of the output is counter-intuitive. All the functions in the package make the cache the second output. Can you please flip the order?

src/transforms/levels.jl
```diff
@@ -0,0 +1,68 @@
...
+"""
+struct Levels{K} <: Stateless
+  levelspec::K
+  ordered::Vector{Symbol}
+end
+
+Levels(pairs::Pair{Symbol}...; ordered=Symbol[]) =
+  Levels(NamedTuple(pairs), ordered)
+
+Levels(pairs::Pair{K}...; ordered=K[]) where {K <: AbstractString} =
```
juliohm, line 23:
```suggestion
Levels(pairs::Pair{K}...; ordered=K[]) where {K<:AbstractString} =
```

src/transforms/levels.jl
```diff
@@ -0,0 +1,67 @@
...
+
+isrevertible(transform::Levels) = true
+
+# when the col is already a categorical array and wanna change levels and order
+categorify(l::AbstractVector, x::CategoricalVector, o) = categorical(x, levels=l, ordered=o), levels(x)
+
+# when the col is normal array and want to change to categorical array
+categorify(l, x::AbstractVector, o) = categorical(x, levels=l, ordered=o), unwrap 
+# when the col is not need for change or convert back to normal array
+categorify(f::Function, x::AbstractVector, o) = o ? (categorical(x, ordered=true), levels(x)) : (f.(x), f)
```
juliohm, line 34:
```suggestion
_categorify(l::AbstractVector, x::CategoricalVector, o) =
  categorical(x, levels=l, ordered=o), levels(x)

# when the col is normal array and want to change to categorical array
_categorify(l::AbstractVector, x::AbstractVector, o) =
  categorical(x, levels=l, ordered=o), unwrap 

# when the col is not need for change or convert back to normal array
_categorify(f::Function, x::AbstractVector, o) =
  o ? (categorical(x, ordered=true), levels(x)) : (f.(x), f)
```

src/transforms/levels.jl
```diff
@@ -0,0 +1,67 @@
...
+
+function apply(transform::Levels,table)
```
juliohm, line 36:
```suggestion
function apply(transform::Levels, table)
```

src/transforms/levels.jl
```diff
@@ -0,0 +1,67 @@
...
+  cols = Tables.columns(table)
+  names = Tables.columnnames(cols)
+  newlevels = transform.levelspec
```
juliohm, line 39: Notice how you are assigning a variable called levelspec to a variable newlevels. Please pick a single name. Given that this variable is only used one time, I would not even bother creating a new variable inside the function. You can access it directly in the get(transform.levelspec, nm, identity) call.

src/transforms/levels.jl
```diff
@@ -0,0 +1,67 @@
...
+  nres = map(names) do nm
+    x = Tables.getcolumn(cols, nm)
+    l = get(newlevels, nm, identity)
+    o = nm ∈ transform.ordered
+    categorify(l, x, o)
```
juliohm, line 44:
```suggestion
    _categorify(l, x, o)
```

src/transforms/levels.jl
juliohm, line 39:
```suggestion
```

src/transforms/levels.jl
```diff
@@ -0,0 +1,67 @@
...
+  end
+  caches = last.(nres)
```
juliohm, line 46:
```suggestion
  categ = first.(result)
  cache = last.(result)
```

juliohm (changes requested):

src/transforms/levels.jl
```diff
@@ -0,0 +1,72 @@
...
+  cols = Tables.columns(table)
+  names = Tables.columnnames(cols)
+  result = map(names) do nm
+    x = Tables.getcolumn(cols, nm)
+    l = get(transform.levelspec, nm, identity)
+    o = nm ∈ transform.ordered
+    _categorify(l, x, o)
+  end
+  categ = first.(result)
+  cache = last.(result)
```
juliohm, line 50:
```suggestion

  result = map(names) do nm
    x = Tables.getcolumn(cols, nm)
    l = get(transform.levelspec, nm, identity)
    o = nm ∈ transform.ordered
    _categorify(l, x, o)
  end
  
  categ = first.(result)
  cache = last.(result)
```

test/Project.toml
```diff
@@ -16,3 +16,4 @@ TypedTables = "9d95f2ec-7b3d-5a63-8d20-e2491e220bb9"
 [compat]
 GR = "=0.59.0"
 Plots = "=1.22.4"
+CategoricalArrays = "=0.10.6"
```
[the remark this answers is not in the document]
Frank-III: So sorry for that. I've fixed it.

src/transforms/levels.jl
```diff
@@ -0,0 +1,74 @@
...
+
+  𝒯 = (; zip(names, categ)...)
+  newtable = 𝒯 |> Tables.materializer(table)
+  
+  newtable, cache
+end
+
+function revert(transform::Levels, newtable, cache)
+  @assert isrevertible(transform)
+
```
juliohm, line 62:
```suggestion
```

juliohm (approved):

juliohm:
@Frank-III that trick with the `_categorify` function was super neat. You avoided a complicated logic by using the first argument of the function as either a function or a vector of levels 👏🏽 

Now we just need to add an entry for the new transform in the docs, right after the entry for Functional.

Frank-III:
Thanks @juliohm, that means a lot to me. It is my first PR and I made a lot of mistakes, I would be more careful with white-space, naming convention and make my code more concise and professional next time. Looking forward to my next PR. :)

src/transforms/levels.jl
```diff
@@ -0,0 +1,72 @@
...
+struct Levels{K} <: Stateless
+  levelspec::K
+  ordered::Vector{Symbol}
+end
+
+Levels(pairs::Pair{Symbol}...; ordered=Symbol[]) =
+  Levels(NamedTuple(pairs), ordered)
+
+Levels(pairs::Pair{K}...; ordered=K[]) where {K<:AbstractString} =
+  Levels(NamedTuple(Symbol(k) => v for (k,v) in pairs), Symbol.(ordered))
```
eliascarv, line 24: The `Levels` constructor will break in this case:
```julia
julia> Levels(:a => ["yes", "no"], :c => [1, 2, 4], :d => [1, 23, 5, 7], ordered = ["a", "b"])
ERROR: MethodError: no method matching Levels(::NamedTuple{(:a, :c, :d), Tuple{Vector{String}, Vector{Int64}, Vector{Int64}}}, ::Vector{String})
Closest candidates are:
  Levels(::K, ::Vector{Symbol}) where K
```
This is a simple solution to this problem:
```suggestion
Levels(levelspec::NamedTuple, ordered::Vector{K}) where {K<:AbstractString} =
  Levels(levelspec, Symbol.(ordered))

Levels(pairs::Pair{Symbol}...; ordered=Symbol[]) =
  Levels(NamedTuple(pairs), ordered)

Levels(pairs::Pair{K}...; ordered=K[]) where {K<:AbstractString} =
  Levels(NamedTuple(Symbol(k) => v for (k,v) in pairs), ordered)
```

Frank-III: Yes, but I guess users would be consistent if they use Symbol in the first place, they would do so for the second one. Do you think it would be better to add the functionality to support it because I don't think to mix `Symbol` and `String` in one expression is good in any use case. Could you talk more about it? (The suggestion is very nice to me since I am so struggled with constructors)

eliascarv: This change is just to make the `Levels` constructor more generic, as in some cases this may be necessary.

juliohm: I agree with @Frank-III that the use case with mixed types is not very welcome. @eliascarv can we simply leave the PR as is without this fuzz with mixed Symbol/String arguments? We need to think in terms of what we want to support and maintain in the long-term. I don't think we want to maintain mixed types in this case, right?

eliascarv: Note that this works:
```julia
Levels("a" => ["yes", "no"], "c" => [1, 2, 4], "d" => [1, 23, 5, 7], ordered = [:a, :b])
```
So it's weird that the opposite doesn't work.

juliohm: Yes, the fact that it works as a side effect of the current implementation doesn't mean that we want to support it officially. If you intentionally write constructors to support mixed types, that means that in future releases we will need to submit more fixes to other transforms that operate with Symbol/String in the same function call.

juliohm:
Thank you again @Frank-III for this great addition to the toolkit :) I will go ahead and merge this following our discussions above. We do *not* want to support mixed Symbol/String types in the same function call.

PR #84, opened 2022-05-26
Add `ndim` option to `EigenAnalysis`

### LINKED ISSUE #17: Add dim option to EigenAnalysis to reduce the number of dimensions

eliascarv:
This PR adds the `ndim` keyword argument to `EigenAnalysis`, it represents the number of dimensions of the output.

closes #17


eliascarv:
@juliohm I changed the `eigenmatrices` function to adapt it to the new `ndim` argument. Please verify the implementation is correct.

src/transforms/eigenanalysis.jl
```diff
@@ -54,38 +60,40 @@ function apply(transform::EigenAnalysis, table)
...
   # original columns names
   cols = Tables.columns(table)
-  names = Tables.columnnames(cols)
+  onames = Tables.columnnames(cols)
 
   # table as matrix
   X = Tables.matrix(table)
 
+  # output dim
+  ndim = _ndim(transform.ndim, X)
```
juliohm, line 69:
```suggestion
  # output dimension
  d = _ndim(transform.ndim, X)
```

src/transforms/eigenanalysis.jl
```diff
@@ -54,38 +60,40 @@ function apply(transform::EigenAnalysis, table)
...
+
   # center the data
   μ = mean(X, dims=1)
   Y = X .- μ
 
   # eigenanalysis of covariance
-  S, S⁻¹ = eigenmatrices(transform, Y)
+  S, S⁻¹ = eigenmatrices(transform, Y, ndim)
```
juliohm, line 76:
```suggestion
  S, S⁻¹ = eigenmatrices(transform, Y, d)
```

src/transforms/eigenanalysis.jl
```diff
@@ -54,38 +60,40 @@ function apply(transform::EigenAnalysis, table)
...
 
   # project the data
   Z = Y * S
 
+  # column names
+  names = Symbol.(:PC, 1:ndim)
```
juliohm, line 82:
```suggestion
  names = Symbol.(:PC, 1:d)
```

src/transforms/eigenanalysis.jl
```diff
@@ -104,75 +112,98 @@ function reapply(transform::EigenAnalysis, table, cache)
...
 
-  # original columns names
-  cols = Tables.columns(table)
-  names = Tables.columnnames(cols)
-
   # table as matrix
   X = Tables.matrix(table)
 
+  # output dim
+  ndim = _ndim(transform.ndim, X)
```
juliohm, line 119:
```suggestion
  # output dimension
  d = _ndim(transform.ndim, X)
```

src/transforms/eigenanalysis.jl
```diff
@@ -104,75 +112,98 @@ function reapply(transform::EigenAnalysis, table, cache)
...
+  μ, S, S⁻¹, onames = cache
 
   # center the data
   Y = X .- μ
 
   # project the data
   Z = Y * S
 
+  # column names
+  names = Symbol.(:PC, 1:ndim)
```
juliohm, line 131:
```suggestion
  names = Symbol.(:PC, 1:d)
```

src/transforms/eigenanalysis.jl
```diff
@@ -104,75 +112,98 @@ function reapply(transform::EigenAnalysis, table, cache)
...
+
   # table with transformed columns
   𝒯 = (; zip(names, eachcol(Z))...)
   𝒯 |> Tables.materializer(table)
 end
 
-function eigenmatrices(transform, Y)
+function eigenmatrices(transform, Y, ndim)
```
juliohm, line 138:
```suggestion
function eigenmatrices(transform, Y, d)
```

src/transforms/eigenanalysis.jl
```diff
@@ -104,75 +112,98 @@ function reapply(transform::EigenAnalysis, table, cache)
...
+    P = V * inv(Λ)
   elseif proj == :VDV
-    Λ   = Diagonal(sqrt.(λ))
-    S   = V * inv(Λ) * transpose(V)
-    S⁻¹ = V * Λ * transpose(V)
+    Λ = Diagonal(sqrt.(λ))
+    P = V * inv(Λ) * transpose(V)
   end
 
+  Pp  = P[:, end:-1:1]
```
juliohm, line 154: Instead of reversing the matrix here, try to use the sortby option in eigen to order the eigenvalues in descending order.

test/transforms.jl
```diff
@@ -1356,6 +1356,37 @@
...
+
+    # ndim
+    x = rand(Normal(), 1500)
+    y = rand(Normal(), 1500)
+    z = rand(Normal(), 1500)
+    t = Table(; x, y, z)
+
+    # PCA
+    T = PCA(2)
+    n, c = apply(T, t)
```
juliohm, line 1368: Here should test that the table n has 2 columns PC1 and PC2

test/transforms.jl
```diff
@@ -1356,6 +1356,37 @@
...
+    Σ = cov(Tables.matrix(n))
+    @test isapprox(Σ[1,2], 0; atol=1e-6)
+    @test isapprox(Σ[2,1], 0; atol=1e-6)
+
+    # DRS
+    T = DRS(2)
+    n, c = apply(T, t)
```
juliohm, line 1375: Same here for DRS

test/transforms.jl
```diff
@@ -1356,6 +1356,37 @@
...
+    Σ = cov(Tables.matrix(n))
+    @test isapprox(Σ[1,2], 0; atol=1e-6)
+    @test isapprox(Σ[2,1], 0; atol=1e-6)
+    @test isapprox(Σ[1,1], 1; atol=1e-6)
+    @test isapprox(Σ[2,2], 1; atol=1e-6)
+
+    # SDS
+    T = SDS(2)
+    n, c = apply(T, t)
```
juliohm, line 1384: And for SDS

juliohm (changes requested):
Much better PR compared to the previous one 💯

src/transforms/eigenanalysis.jl
eliascarv: If eigenvalues ​​and eigenvectors are reversed, the output will be slightly different, breaking compatibility with the current reference test.

juliohm: I think that is fine. We just need to review the results carefully and add this change as a minor release.

juliohm: I didn't know we had tests that depended on the order of the vectors in the projection BTW.

eliascarv: It's not just a change in the x and y order, it's a change in the output values.

juliohm: What change exactly? Can you please clarify what tests fail if we flip the order of vectors?

eliascarv: The S matrix will be slightly different because `inv(Λ)` will return a slightly different matrix.

eliascarv: That's why I had to change the reference tests in the previous PR.

juliohm: > The `S` matrix will be slightly different because `inv(Λ)` will return a slightly different matrix.

That was not my question, but the `inv(Λ)` expression before and after the proposed change should be the same except for the order of the entries in the diagonal. This is a diagonal matrix.

Can you please share which line in the test suite fails when you just flip the order without discarding vectors from the basis?

eliascarv: In visual tests, the plots will be different if I change the order of the eigenvalues.

src/transforms/eigenanalysis.jl
```diff
@@ -104,32 +112,36 @@ function reapply(transform::EigenAnalysis, table, cache)
...
 
-function eigenmatrices(transform, Y)
+function eigenmatrices(transform, Y, d)
   proj = transform.proj
 
   Σ = cov(Y)
-  λ, V = eigen(Σ)
+  F = eigen(Σ)
+  λ = F.values[end:-1:1]
+  V = F.vectors[:, end:-1:1]
```
juliohm, line 144: Please use the sortby option of eigen instead and preserve the previous version with two output arguments.

src/transforms/eigenanalysis.jl
```diff
@@ -144,35 +156,56 @@ function eigenmatrices(transform, Y)
     S⁻¹ = V * Λ * transpose(V)
   end
 
-  S, S⁻¹
+  S[:, 1:d], S⁻¹[1:d, :]
```
juliohm, line 159: I am assuming you double checked carefully that the 1:d in the S^-1 case appears in the rows instead of in the columns.

test/transforms.jl
```diff
@@ -1333,12 +1333,12 @@
...
-      p₄ = scatter(t₄.x, t₄.y, label="VDV")
-      p₅ = scatter(t₅.x, t₅.y, label="PCA")
-      p₆ = scatter(t₆.x, t₆.y, label="DRS")
-      p₇ = scatter(t₇.x, t₇.y, label="SDS")
+      p₂ = scatter(t₂.PC2, t₂.PC1, label="V")
+      p₃ = scatter(t₃.PC2, t₃.PC1, label="VD")
+      p₄ = scatter(t₄.PC1, t₄.PC2, label="VDV")
+      p₅ = scatter(t₅.PC2, t₅.PC1, label="PCA")
+      p₆ = scatter(t₆.PC2, t₆.PC1, label="DRS")
+      p₇ = scatter(t₇.PC1, t₇.PC2, label="SDS")
```
juliohm, line 1341: These tests should all show PC1 and PC2. Please update the plot file instead.

juliohm (changes requested):
The visual test file needs to be updated.

src/transforms/eigenanalysis.jl
eliascarv: It is not possible to reverse the order using the sortby kwarg and the eigen function does not have `rev` kwarg like the sort function.

juliohm: Please read the docs of eigen, you can pass a sortby function. `eigen(A, sortby=l -> -real(l))` for example.

test/transforms.jl
```diff
@@ -1356,6 +1356,40 @@
...
+    z = rand(Normal(), 1500)
+    t = Table(; x, y, z)
+
+    # PCA
+    T = PCA(2)
+    n, c = apply(T, t)
+    Σ = cov(Tables.matrix(n))
+    @test isapprox(Σ[1,2], 0; atol=1e-6)
+    @test isapprox(Σ[2,1], 0; atol=1e-6)
+    @test Tables.columnnames(n) == (:PC1, :PC2)
```
juliohm, line 1372:
```suggestion
    @test Tables.columnnames(n) == (:PC1, :PC2)
    @test isapprox(Σ[1,2], 0; atol=1e-6)
    @test isapprox(Σ[2,1], 0; atol=1e-6)
```

test/transforms.jl
```diff
@@ -1356,6 +1356,40 @@
...
+
+    # DRS
+    T = DRS(2)
+    n, c = apply(T, t)
+    Σ = cov(Tables.matrix(n))
+    @test isapprox(Σ[1,2], 0; atol=1e-6)
+    @test isapprox(Σ[2,1], 0; atol=1e-6)
+    @test isapprox(Σ[1,1], 1; atol=1e-6)
+    @test isapprox(Σ[2,2], 1; atol=1e-6)
+    @test Tables.columnnames(n) == (:PC1, :PC2)
```
juliohm, line 1382:
```suggestion
    @test Tables.columnnames(n) == (:PC1, :PC2)
    @test isapprox(Σ[1,2], 0; atol=1e-6)
    @test isapprox(Σ[2,1], 0; atol=1e-6)
    @test isapprox(Σ[1,1], 1; atol=1e-6)
    @test isapprox(Σ[2,2], 1; atol=1e-6)
```

test/transforms.jl
```diff
@@ -1356,6 +1356,40 @@
...
+
+    # SDS
+    T = SDS(2)
+    n, c = apply(T, t)
+    Σ = cov(Tables.matrix(n))
+    @test isapprox(Σ[1,2], 0; atol=1e-6)
+    @test isapprox(Σ[2,1], 0; atol=1e-6)
+    @test isapprox(Σ[1,1], 1; atol=1e-6)
+    @test isapprox(Σ[2,2], 1; atol=1e-6)
+    @test Tables.columnnames(n) == (:PC1, :PC2)
```
juliohm, line 1392:
```suggestion
    @test Tables.columnnames(n) == (:PC1, :PC2)
    @test isapprox(Σ[1,2], 0; atol=1e-6)
    @test isapprox(Σ[2,1], 0; atol=1e-6)
    @test isapprox(Σ[1,1], 1; atol=1e-6)
    @test isapprox(Σ[2,2], 1; atol=1e-6)
```

test/transforms.jl
```diff
@@ -1356,6 +1356,40 @@
     @test Tables.isrowtable(n)
     rtₒ = revert(T, n, c)
     @test Tables.matrix(rt) ≈ Tables.matrix(rtₒ)
+
+    # ndim
+    x = rand(Normal(), 1500)
+    y = rand(Normal(), 1500)
+    z = rand(Normal(), 1500)
```
juliohm, line 1363: Please create a more challenging test, these 3 variables are uncorrelated already even before the transformation. Create x = randn(1000) and then y = x + randn(1000)

juliohm (changes requested):
The data set used in the new test is too trivial.

test/transforms.jl
eliascarv: And z? x * y?

juliohm (approved):

juliohm:
Nice work @eliascarv and great linear algebra exercise as well 👍🏽

Issue #74, opened 2022-05-10
Sort


juliohm:
Add a Sort transform that sorts the rows a la `sort` from Base. The transform should allow the specification of columns, ascending vs descending order, etc.

juliohm:
Fixed on master.

PR #87, opened 2022-05-30
Add sort transform

ClaroHenrique:
This PR aims to add the `sort` transform mentioned in [issue 74](https://github.com/JuliaML/TableTransforms.jl/issues/74)
This transform returns the table with rows sorted by values of a specific column in ascending or descending order,


src/transforms/sort.jl
```diff
@@ -0,0 +1,54 @@
...
+isrevertible(::Type{<:Sort}) = true
+
+function apply(transform::Sort, table)
+  # use selected column to calculate new order
+  scol = Tables.getcolumn(table, transform.by)
+  neworder = sortperm(scol, rev=!transform.ascending)
+
+  # sort rows
+  rows = Tables.rowtable(table)
+  rows = rows[neworder]
```
juliohm, line 39: Nice I think this is a great approach. We need to collect the rows anyways so using Tables.rowtable seems to make sense. I think Tables.rows is lazy sometimes. If it is not lazy, we should use it instead to use the most efficient representation of rows for the given table type.

juliohm (changes requested):
This is an awesome PR @ClaroHenrique , thanks for the contribution. I left a few review comments so that it can be made more consistent with the Base.sort API.

src/transforms/sort.jl
```diff
@@ -0,0 +1,54 @@
+# ------------------------------------------------------------------
+# Licensed under the MIT License. See LICENSE in the project root.
+# ------------------------------------------------------------------
+
+"""
+    Sort(by, ascending)
+
+Returns a table copy with rows sorted by values of a specific column.
+The `by` value specifies the column used to sort, it can be a index (Int) or a name (Symbol).
+The `ascending` value is a boolean that specifies the sort order, the default value is true.
```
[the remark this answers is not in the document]
ClaroHenrique: makes sense

src/transforms/sort.jl
```diff
@@ -0,0 +1,54 @@
...
+  ascending::Bool
+end
+
+Sort(by; ascending=true) = Sort(by, ascending)
+
+isrevertible(::Type{<:Sort}) = true
+
+function apply(transform::Sort, table)
+  # use selected column to calculate new order
+  scol = Tables.getcolumn(table, transform.by)
```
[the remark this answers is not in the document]
ClaroHenrique: ok

src/transforms/sort.jl
```diff
@@ -0,0 +1,54 @@
+# ------------------------------------------------------------------
+# Licensed under the MIT License. See LICENSE in the project root.
+# ------------------------------------------------------------------
+
+"""
+    Sort(by, ascending)
```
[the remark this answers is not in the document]
juliohm: Maybe just `col` is enough. The docstring can refer to it and provide meaning 👍🏽

src/transforms/sort.jl
ClaroHenrique: I dind't find in documentation specifically if its lazy or not. There is a way to check in code?

juliohm: Good question. I don't think we have a mechanism to discover the lazyness programmatically, as far as I remember. If the docstring of Table.rows doesn't require the AbstractArray interface then we probably can't use it.

If Table.rows requires the AbstractArray interface, that means that we can use getindex (i.e. A[inds]) to reorder the array as you did. If it does not require the interface, that means that someone may implement a lazy Table.rows and our code will break, in which case it is safer to always use Tables.rowtable as you implemented already.

src/transforms/sort.jl
```diff
@@ -0,0 +1,52 @@
...
+  cols = Tables.columns(table)
+  scol = Tables.getcolumn(cols, transform.col)
+  neworder = sortperm(scol, rev=transform.rev)
+
+  # sort rows
+  rows = Tables.rowtable(table)
+  rows = rows[neworder]
+
+  newtable = rows |> Tables.materializer(table)
+  newtable, neworder
```
juliohm, line 40:
```suggestion
  inds = sortperm(scol, rev=transform.rev)

  # sort rows
  rows = Tables.rowtable(table)
  rows = rows[inds]

  newtable = rows |> Tables.materializer(table)
  newtable, inds
```

src/transforms/sort.jl
```diff
@@ -0,0 +1,52 @@
...
+
+function revert(::Sort, newtable, cache)
+  # use cache to recalculate old order
+  oldorder = sortperm(cache)
+
+  # undo rows sort
+  rows = Tables.rowtable(newtable)
+  rows = rows[oldorder]
+
+  rows |> Tables.materializer(newtable)
```
juliohm, line 51:
```suggestion
  inds = sortperm(cache)

  # undo rows sort
  rows = Tables.rowtable(newtable)
  rows = rows[inds]

  rows |> Tables.materializer(newtable)
```

juliohm (changes requested):

juliohm (approved):

juliohm:
@ClaroHenrique that is looking great already. Can you please add the transform to the docs page in the sample place after the Levels transform?

src/transforms/sort.jl
```diff
@@ -0,0 +1,52 @@
+# ------------------------------------------------------------------
+# Licensed under the MIT License. See LICENSE in the project root.
+# ------------------------------------------------------------------
+
+"""
+    Sort(col, rev)
```
eliascarv, line 6:
```suggestion
    Sort(col; rev=false)
```
Only that. The rest of the code is perfect!

eliascarv (approved):

juliohm:
@ClaroHenrique I was thinking if we could generalize this PR a bit and instead of just providing the `rev` keyword argument, actually forward all keyword arguments to the internal `sortperm` call?

```julia
sortperm(v; alg::Algorithm=DEFAULT_UNSTABLE, lt=isless, by=identity, rev::Bool=false, order::Ordering=Forward)
```

That would be a much richer set of options for end users. Please let me know if you have questions about how to forward keywords generically. You can simply type `foo(; kwargs...) = bar(; kwargs)` if I am not mistaken.

juliohm:
Another possible feature that we can consider here is not only sort by a single column, but by a list of columns. In that case whenever a user specifies a ColSpec, we retrieve all columns and treat them as a collection of named tuples to be sorted lexicographically. For example, `Sort([:a,:b])`would sort a vector of the form [(a=1,b=2), (a=2,b=1), ...].

eliascarv:
> @ClaroHenrique I was thinking if we could generalize this PR a bit and instead of just providing the `rev` keyword argument, actually forward all keyword arguments to the internal `sortperm` call?
> 
> ```julia
> sortperm(v; alg::Algorithm=DEFAULT_UNSTABLE, lt=isless, by=identity, rev::Bool=false, order::Ordering=Forward)
> ```
> 
> That would be a much richer set of options for end users. Please let me know if you have questions about how to forward keywords generically. You can simply type `foo(; kwargs...) = bar(; kwargs)` if I am not mistaken.

Like this:
```julia
struct Sort{T} <: Stateless
  col::Symbol
  kwargs::T
end

Sort(col::Symbol; kwargs...) = Sort(col, value(kwargs))
```

ClaroHenrique:
Ok @eliascarv, thanks.
Are going to work on the multi column sorting also?

src/transforms/sort.jl
```diff
@@ -0,0 +1,74 @@
+# ------------------------------------------------------------------
+# Licensed under the MIT License. See LICENSE in the project root.
+# ------------------------------------------------------------------
+
+"""
+    Sort(col₁, col₂, ..., colₙ; alg=DEFAULT_UNSTABLE, lt=isless, by=identity, rev=false, order=Forward)
+    Sort([col₁, col₂, ..., colₙ]; alg=DEFAULT_UNSTABLE, lt=isless, by=identity, rev=false, order=Forward)
+    Sort((col₁, col₂, ..., colₙ); alg=DEFAULT_UNSTABLE, lt=isless, by=identity, rev=false, order=Forward)
```
juliohm, line 8: We just need to write `kwargs` in the docstring and mention that they are forwarded to `sortperm`. No need to hard-coding the options here. `sortperm` may support more options in the future.

src/transforms/sort.jl
```diff
@@ -0,0 +1,74 @@
...
+
+Sort the table by the values of the selected columns `col₁`, `col₂`, ..., `colₙ`.
+The keyword arguments are the same as in the `sort` function.
```
juliohm, line 11: Here it should be sortperm instead of sort.

src/transforms/sort.jl
```diff
@@ -0,0 +1,74 @@
...
+Sort(:a)
+Sort(:a, :c, rev=true)
+Sort([1, 3, 5], by=row -> abs.(row))
+Sort(("a", "c", "e"))
+Sort(r"[ace]")
+```
+
+## Notes
+
+* The row passed to `by` kwarg is the selection row, not the table row.
```
juliohm, line 29: I don't understand this comment. The `by` keyword is usually a function that applies to the entries before the sorting, correct?

juliohm (changes requested):
Great additions @eliascarv and @ClaroHenrique , please check this last round of review comments.

juliohm (approved):
Awesome work. Let's just wait for the tests to pass. I will squash and merge the commits afterwards 🎉

PR #91, opened 2022-06-08
Refactor Levels Tranform

eliascarv:


src/transforms/levels.jl
```diff
@@ -1,72 +1,76 @@
...
+# ------------------------------------------------------------------
+
 """
-    Levels(:a => ["yes", "no"])
+    Levels(col₁ => levels₁, col₂ => levels₂, ..., colₙ => levelsₙ; ordered=nothing)
 
-Return a copy of the table with specified levels and orders for categorical columns
-allowing only changing the order of the column.
+Make selected columns `col₁`, `col₂`, ..., `colₙ` categorical by specifying the levels of each one.
+`ordered` keyword argument can be a tuple, vector or regex that selects the columns where `ordered=true`.
```
juliohm, line 9:
```suggestion
Convert columns `col₁`, `col₂`, ..., `colₙ` to categorical arrays with given levels `levels₁`, `levels₂`, ..., `levelsₙ`.
Optionally, specify which columns are `ordered`.
```

test/transforms.jl
```diff
@@ -1277,60 +1277,88 @@
   end
   
   @testset "Levels" begin
-    a = categorical(["yes", "no", "no", "no", "yes"]) 
-    b = categorical([1, 2, 4, 2, 8], ordered=false) 
-    c = categorical([1, 2, 1, 2, 1]) 
-    d = [1, 23, 5, 7, 7]
-    e = categorical([2, 3, 1, 4, 1])
-    t = Table(; a, b, c, d, e)
```
juliohm, line 1285: Please do not change existing tests unless strictly necessary. Your changes should be backward compatible right? You are changing the input tables for the tests, which requires a closer look and review. If you can undo these changes I can more easily review the additions being certain that nothing that used to work is breaking.

juliohm (changes requested):
Please undo changes to existing tests and add new tests. I understand that the PR is adding new functionality, not breaking existing functionality.

test/transforms.jl
eliascarv: I added and updated the old tests. I removed tests that checked for incorrect behavior of the `ordered` keyword and added more tests.

PR #92, opened 2022-06-09
Add OneHot Transform

### LINKED ISSUE #72: OneHotEncoding
The transform takes categorical columns and converts into multiple columns with 1s and 0s using the CategoricalArrays.jl API.

eliascarv:
This PR implements OneHot transform.

closes #72


src/TableTransforms.jl
```diff
@@ -38,6 +38,7 @@ export
   Coerce,
   RowTable,
   ColTable,
+  OneHotEncoding,
```
juliohm, line 41: I wonder if we could adopt the shorter name OneHot? It should be clear and more concise. What do you think?

src/transforms/onehotencoding.jl
```diff
@@ -0,0 +1,86 @@
...
+  @assert col ∈ names "Invalid column."
+  return col
+end
+
+_name(nm, names) = 
+  nm ∈ names ? _name(Symbol("$(nm)_"), names) : nm
+
+function apply(transform::OneHotEncoding, table)
+  cols = Tables.columns(table)
+  names = collect(Tables.columnnames(cols))
```
juliohm, line 39: Do we really need to collect the names? Isn't Tables.columnnames always indexable?

src/transforms/onehotencoding.jl
```diff
@@ -0,0 +1,86 @@
...
+  columns = AbstractVector[Tables.getcolumn(cols, nm) for nm in names]
```
juliohm, line 40: Is this related to the issue we had in other transform where we are avoiding specialization of type in order to revert later? Why do we need the AbstractVector here? Maybe we should use Any as we did in the other case?

src/transforms/onehotencoding.jl
```diff
@@ -0,0 +1,86 @@
...
+  
+  name = _colname(transform.col, names)
+  ind = findfirst(==(name), names)
+  x = columns[ind]
+
+  if !isa(x, CategoricalArray)
+    throw(ArgumentError("The $name column must be cetegorical."))
+  end
```
juliohm, line 48: This here should be a new assertion function following our assertion function api. I think we only have assert_continuous at the moment, we should add assert_categorical

src/transforms/onehotencoding.jl
```diff
@@ -0,0 +1,86 @@
...
+
+  xl = levels(x)
+  onehot = map(xl) do l
+    nm = Symbol("$(name)_$l")
+    nm = _name(nm, names)
+    nm, x .== l
+  end
+
+  newcols, newnms = last.(onehot), first.(onehot)
```
juliohm, line 57: Can you please swap the order here? Why you create a pair with last/first instead of first/last which is more natural? Also, we always think of these pairs as names => cols, not cols => names. Let's try to follow this pattern.

src/transforms/onehotencoding.jl
```diff
@@ -0,0 +1,86 @@
...
+
+  splice!(columns, ind, newcols)
+  splice!(names, ind, newnms)
```
juliohm, line 60: splice! is a wonderful funciton, I need to use it more!

juliohm: Please reorder these lines so that the names => cols pattern is respected.

src/transforms/onehotencoding.jl
```diff
@@ -0,0 +1,86 @@
...
+
+  𝒯 = (; zip(names, columns)...)
+  newtable = 𝒯 |> Tables.materializer(table)
+  newtable, (name, inds, xl, isordered(x))
+end
+
+function revert(::OneHotEncoding, newtable, cache)
+  cols = Tables.columns(newtable)
+  names = collect(Tables.columnnames(cols))
+  columns = AbstractVector[Tables.getcolumn(cols, nm) for nm in names]
```
juliohm, line 72: Again I wonder why we need to annotate the AbstractVector and if we could just use Any instead

juliohm (changes requested):
Nice usage of splice!, that function is super helpful.

Attached a couple of review suggestions. I think we can improve the code a bit further.

src/transforms/onehotencoding.jl
eliascarv: Yes, the `splice!` function needs a Vector as the first argument.

juliohm: I would call collect or vec in the splice! function then. Calling it here too early makes the reader of the code wonder why this is necessary.

src/transforms/onehotencoding.jl
eliascarv: So I think Any might be a better choice.

src/TableTransforms.jl
eliascarv: I think that's a good name.

src/transforms/onehotencoding.jl
juliohm: what about splice!(collect(names), ind, newnms)?

eliascarv: The `splice!` function returns the removed value, not the transformed vector. So I need a vector in a variable.

src/assertions.jl
```diff
@@ -7,3 +7,8 @@ function assert_continuous(table)
   types = schema(table).scitypes
   @assert all(T <: Continuous for T in types) "columns must hold continuous variables"
 end
+
+# assert that column is categorical
+function assert_categorical(x)
+  @assert x isa CategoricalArray "The selected column must be cetegorical."
```
juliohm, line 13: We need to use the scitype of the column instead. Check the assert_continuous for an idea. The return of the function will be something like Categorical. Check the type hierarchy of ScientificTypes.jl to learn more.

src/transforms/onehot.jl
```diff
@@ -0,0 +1,84 @@
...
+isrevertible(::Type{<:OneHot}) = true
+
+_colname(col::Integer, names) = names[col]
+_colname(col::AbstractString, names) =
+  _colname(Symbol(col), names)
+
+function _colname(col::Symbol, names)
+  @assert col ∈ names "Invalid column selection."
+  return col
+end
```
juliohm, line 32: So you are reproducing the logic of the ColSpec feature here right? Maybe we can reuse the ColSpec functionality but restrict at construction time what ColSpec are allowed? In this case we want to restrict ourselves to a ColSpec that consist of a single column, right?

src/transforms/onehot.jl
```diff
@@ -0,0 +1,84 @@
...
+
+_name(nm, names) = 
+  nm ∈ names ? _name(Symbol("$(nm)_"), names) : nm
```
juliohm, line 35: So this function is appending an underscore when the column name already exists right? Can we get rid of it and then inline the implementation inside the apply function? You still have this tendency to create tiny isolated functions that could live in apply :)

src/transforms/onehot.jl
```diff
@@ -0,0 +1,84 @@
...
+
+function apply(transform::OneHot, table)
+  cols = Tables.columns(table)
+  names = collect(Tables.columnnames(cols))
```
juliohm, line 39:
```suggestion
  names = Tables.columnnames(cols) |> collect
```

juliohm (changes requested):
The PR is looking great already, just a few improvements before we can merge it.

src/assertions.jl
```diff
@@ -7,3 +7,9 @@ function assert_continuous(table)
   types = schema(table).scitypes
   @assert all(T <: Continuous for T in types) "columns must hold continuous variables"
 end
+
+# assert that column is categorical
+function assert_categorical(x)
+  T = elscitype(x)
+  @assert T <: Finite "The selected column must be cetegorical."
```
juliohm, line 14:
```suggestion
  @assert elscitype(x) <: Finite "The selected column must be categorical."
```

juliohm (commented):
Awesome, will merge it soon after the tests pass.

juliohm (approved):

PR #97, opened 2022-06-12
Add Sample

eliascarv:


src/TableTransforms.jl
```diff
@@ -39,6 +41,7 @@ export
   Coerce,
   Levels,
   OneHot,
+  Sample,
```
juliohm, line 44: Let's move this to after Sort

src/TableTransforms.jl
```diff
@@ -12,6 +12,8 @@ using LinearAlgebra
 using Statistics
 using PrettyTables
 using CategoricalArrays
+using Random
```
juliohm, line 15:
```suggestion
using Random: AbstractRNG
```

src/TableTransforms.jl
```diff
@@ -12,6 +12,8 @@ using LinearAlgebra
...
+using StatsBase
```
juliohm, line 16:
```suggestion
using StatsBase: AbstractWeights
```
Also, please move this line to before LinearAlgebra given that it is not a stdlib.

src/transforms/sample.jl
```diff
@@ -0,0 +1,72 @@
...
+  Sample(rng, wv, n, replace, ordered)
+
+Sample(rng::AbstractRNG, n::Int; replace=true, ordered=false) =
+  Sample(rng, nothing, n, replace, ordered)
+
+Sample(wv::AbstractWeights, n::Int; replace=true, ordered=false) =
+  Sample(Random.GLOBAL_RNG, wv, n, replace, ordered)
+
+Sample(n::Int; replace=true, ordered=false) =
+  Sample(Random.GLOBAL_RNG, nothing, n, replace, ordered)
```
juliohm, line 48: I wonder if we could just create a Sample(args; kwargs) and forward the args and kwargs to sample. That would make the code less verbose and less dependent on AbstractRNG and AbstractWeights.

juliohm: Maybe that is not possible because we need to pass the rows as the second argument to the function sample right?

src/transforms/sample.jl
```diff
@@ -0,0 +1,72 @@
...
+  replace = transform.replace
+  ordered = transform.ordered
+
+  if isnothing(wv)
+    newrows = sample(rng, rows, n; replace, ordered)
+  else
+    newrows = sample(rng, rows, wv, n; replace, ordered)
+  end
+
+  newtable = newrows |> Tables.materializer(table)
```
juliohm, line 67:
```suggestion
  newtable = newrows |> Tables.materializer(table)
  
```

juliohm (changes requested):
Another great PR. We are coding with the same codestyle now basically, which is super nice.

src/transforms/sample.jl
eliascarv: Yes, and in each method, the Sample builder has different default values. Also, type annotations bring better error messages.

juliohm (approved):
Awesome 💯

PR #100, opened 2022-06-13
Add shows for: Transform, Sequential and Parallel

eliascarv:


src/transforms/parallel.jl
```diff
@@ -18,6 +18,21 @@ struct Parallel <: Transform
   transforms::Vector{Transform}
 end
 
+function Base.show(io::IO, transform::Parallel)
+  print(io, transform.transforms[1])
+  for t in transform.transforms[2:end]
+    print(io, " ⊔ $t")
+  end
```
juliohm, line 25: I think we can use the 3-argument version of join to create the final string and then use a single println.

juliohm: Also, what about showing parallel transforms as a stack of transforms? I wonder if we could leverage some of the pretty printing packages to create more interesting visuals for these pipelines.

src/transforms.jl
```diff
@@ -113,6 +113,21 @@ isrevertible(::Type{<:Transform}) = false
...
+  print(io, "$(nameof(T))($(join(values, ", ")))")
+end
+
+function Base.show(io::IO, ::MIME"text/plain", transform::Transform)
+  T = typeof(transform)
+  print(io, "$(nameof(T)) transform with parameters:")
+  for field in fieldnames(T)
+    print(io, "\n $field = $(getfield(transform, field))")
+  end
+end
```
juliohm, line 129: So these generic methods could go in a separate PR given that this PR is focused on Sequential and Parallel.

juliohm (changes requested):
We also need tests when the final implementation is ready.

juliohm:
I wonder if we could leverage the printing functions of AbstractTrees.jl to achieve something like:

image

That is, create this tree that looks like a pipeline.

test/shows.jl
```diff
@@ -0,0 +1,68 @@
...
+  @testset "Sequential" begin
+    t1 = Select(:x, :z)
+    t2 = ZScore()
+    t3 = Scale(low=0, high=1)
+    pipeline = t1 → t2 → t3
+
+    # compact mode
+    show(io, pipeline)
+    iostr = String(take!(io))
+    @test iostr == "Select{Tuple{Symbol, Symbol}}((:x, :z)) → ZScore() → Scale{Int64}(0, 1)"
```
juliohm, line 12: I think there is a shorter code for testing these prints: https://github.com/JuliaEarth/GeoStatsBase.jl/blob/d43cb39486a3f9d87d68276a0366f26774689de0/test/ensembles.jl#L28

src/transforms/parallel.jl
```diff
@@ -18,6 +18,18 @@ struct Parallel <: Transform
...
+# AbstractTrees interface
+AbstractTrees.nodevalue(::Parallel) = Parallel
+AbstractTrees.children(p::Parallel) = p.transforms
+
+Base.show(io::IO, p::Parallel) =
+  print(io, join(p.transforms, " ⊔ "))
+
+function Base.show(io::IO, ::MIME"text/plain", p::Parallel)
+  tree = repr_tree(p, context=io)
+  print(io, tree[begin:end-1]) # remove \n at end
```
juliohm, line 30: Can't we use print_tree directly from AbstractTrees.jl? Does it handle the trailing \n for us maybe?

juliohm (approved):

PR #101, opened 2022-06-15
Add Generic Shows

eliascarv:
Examples:
```julia
julia> EigenAnalysis(:VDV)
EigenAnalysis transform
├─ proj = :VDV
├─ maxdim = nothing
└─ pratio = 1.0

julia> EigenAnalysis(:VDV) |> println
EigenAnalysis(:VDV, nothing, 1.0)

julia> Scale(low=0, high=1)
Scale transform
├─ low = 0
└─ high = 1

julia> Scale(low=0, high=1) |> println
Scale(0, 1)

julia> begin
           f1 = ZScore()
           f2 = Scale()
           f3 = Functional(cos)
           f4 = MinMax()
           pipeline = (f1 → f2) ⊔ (f3 → f4)
       end
Parallel
├─ Sequential
│  ├─ ZScore()
│  └─ Scale(0.25, 0.75)
└─ Sequential
   ├─ Functional(cos)
   └─ Scale(0, 1)

julia> pipeline |> println
ZScore() → Scale(0.25, 0.75) ⊔ Functional(cos) → Scale(0, 1)
```

### DIFF
--- a/src/transforms.jl
+++ b/src/transforms.jl
@@ -123,11 +123,15 @@ end
 
 function Base.show(io::IO, ::MIME"text/plain", transform::Transform)
   T = typeof(transform)
-  print(io, "$(nameof(T)) transform with parameters:")
-  for field in fieldnames(T)
+  print(io, "$(nameof(T)) transform")
+  
+  fnames = fieldnames(T)
+  len = length(fnames)
+  for (i, field) in enumerate(fnames)
+    div = i == len ? "\n└─ " : "\n├─ "
     val = getfield(transform, field)
     str = repr(val, context=io)
-    print(io, "\n $field = $str")
+    print(io, "$div$field = $str")
   end
 end
 
--- a/src/transforms.jl
+++ b/src/transforms.jl
@@ -113,6 +113,24 @@ isrevertible(::Type{<:Transform}) = false
 
 (transform::Transform)(table) = apply(transform, table) |> first
 
+# Generic shows
+function Base.show(io::IO, transform::Transform)
+  T = typeof(transform)
+  vals = getfield.(Ref(transform), fieldnames(T))
+  strs = repr.(vals, context=io)
+  print(io, "$(nameof(T))($(join(strs, ", ")))")
+end
+
+function Base.show(io::IO, ::MIME"text/plain", transform::Transform)
+  T = typeof(transform)
+  print(io, "$(nameof(T)) transform with parameters:")
+  for field in fieldnames(T)
+    val = getfield(transform, field)
+    str = repr(val, context=io)
+    print(io, "\n $field = $str")
+  end
+end
+
 # --------------------
 # STATELESS FALLBACKS
 # --------------------
--- a/test/shows.jl
+++ b/test/shows.jl
@@ -7,15 +7,15 @@
 
     # compact mode
     iostr = sprint(show, pipeline)
-    @test iostr == "Select{Tuple{Symbol, Symbol}}((:x, :z)) → ZScore() → Scale{Int64}(0, 1)"
+    @test iostr == "Select((:x, :z)) → ZScore() → Scale(0, 1)"
 
     # full mode
     iostr = sprint(show, MIME("text/plain"), pipeline)
     @test iostr == """
     Sequential
-    ├─ Select{Tuple{Symbol, Symbol}}((:x, :z))
+    ├─ Select((:x, :z))
     ├─ ZScore()
-    └─ Scale{Int64}(0, 1)"""
+    └─ Scale(0, 1)"""
   end
 
   @testset "Parallel" begin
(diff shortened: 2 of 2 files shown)


src/transforms.jl
```diff
@@ -113,6 +113,24 @@ isrevertible(::Type{<:Transform}) = false
...
+function Base.show(io::IO, transform::Transform)
+  T = typeof(transform)
+  vals = getfield.(Ref(transform), fieldnames(T))
+  strs = repr.(vals, context=io)
+  print(io, "$(nameof(T))($(join(strs, ", ")))")
+end
+
+function Base.show(io::IO, ::MIME"text/plain", transform::Transform)
+  T = typeof(transform)
+  print(io, "$(nameof(T)) transform with parameters:")
```
juliohm, line 126:
```suggestion
  print(io, "$(nameof(T))")
```

src/transforms.jl
```diff
@@ -113,6 +113,24 @@ isrevertible(::Type{<:Transform}) = false
...
+  for field in fieldnames(T)
+    val = getfield(transform, field)
+    str = repr(val, context=io)
+    print(io, "\n $field = $str")
```
juliohm, line 130:
```suggestion
    print(io, "\n└─ $field = $str")
```

juliohm (changes requested):
Small adjustments to the non-compact version.

ablaom:
Sorry to jump in here but I want to raise a general issue about how best to implement uniform display of objects defined essentially by an interface. I mean, I'm assuming the display you are implementing here can only ever work here for the built-in  transformers that subtype your "optional" abstract `Transformer` type, right? 

We want to get rid of the (compulsory) abstract typing for models in MLJ and will face the same issue. Now, in MLJ, composite models get displayed recursively, to some depth, which allows the user wants to see the nested hyperparameters, up to some depth, which I think is very helpful. Here's an example:

```julia
ProbabilisticPipeline(
  continuous_encoder = ContinuousEncoder(
        drop_last = false, 
        one_hot_ordered_factors = false), 
  feature_selector = FeatureSelector(
        features = [:TechSupport__No, Symbol("PaymentMethod__Credit card (automatic)"), :StreamingMovies__No, Symbol("Contract__One year"), Symbol("Contract__Two year"), :Partner__Yes, :OnlineBackup__No, :Dependents__Yes, :Partner__No, :gender__Male, Symbol("PaymentMethod__Mailed check"), Symbol("OnlineSecurity__No internet service"), :OnlineSecurity__No, Symbol("PaymentMethod__Electronic check"), :OnlineSecurity__Yes, Symbol("InternetService__Fiber optic"), :TechSupport__Yes, Symbol("OnlineBackup__No internet service"), :InternetService__No, :StreamingTV__Yes, :PhoneService__No, :PhoneService__Yes, Symbol("DeviceProtection__No internet service"), Symbol("StreamingTV__No internet service"), Symbol("StreamingMovies__No internet service"), Symbol("TechSupport__No internet service"), :MultipleLines__Yes, Symbol("MultipleLines__No phone service"), :InternetService__DSL], 
        ignore = true), 
  evo_tree_classifier = EvoTreeClassifier(
        loss = EvoTrees.Softmax(), 
        nrounds = 10, 
        λ = 0.0, 
        γ = 0.0, 
        η = 0.1, 
        max_depth = 5, 
        min_weight = 1.0, 
        rowsample = 1.0, 
        colsample = 1.0, 
        nbins = 64, 
        α = 0.5, 
        metric = :mlogloss, 
        rng = Random.MersenneTwister(123, (0, 86172, 85170, 780)), 
        device = "cpu"), 
  cache = true)
```
(Another  thing I like about this display is that it generally matches the way you *construct* the object, as all MLJ models have a keyword constructor.)

So ideally, when MLJ surrenders abstract types as compulsory, we want to retain this display feature. But that means asking client model types to overload our kind of `show` method, or, better, agreeing on some kind of functionality that lives in a shared library. But the `show` that MLJ will want for TableTransforms.jl models is going to different from  what is proposed here. 

To be clear, I am not objecting to the current PR (I wouldn't view pre-1.0 display changes as breaking anyhow). But perhaps this broader context can inform what you decide here, or you may have some suggestion for how we address this issue more widely?

@juliohm

juliohm:
Thank you @ablaom for the feedback. I understand that you are trying to remove the abstract parent type approach from MLJ in future releases (that is great) and that you are concerned that objects won't be displayed in a consistent way when TableTransforms.jl has its own convention for displaying transforms.

>Another thing I like about this display is that it generally matches the way you construct the object, as all MLJ models have a keyword constructor

That is a nice feature to have from a developer's perspective. From the perspective of an end-user, that display is not very friendly. I do believe that a display system has the purpose of clarifying what is happening with well-formatted text, unicode, etc. It is not the purpose of a display system to write valid Julia code. As I see it, the only benefit of having valid Julia code in Base.show is that users can copy/paste the produced output to debug issues in a Julia session. However, debuggers are now available in Julia and are the best method for debugging.

>But that means asking client model types to overload our kind of show method, or, better, agreeing on some kind of functionality that lives in a shared library. But the show that MLJ will want for TableTransforms.jl models is going to different from what is proposed here.

It seems that you are trying to achieve a project-wide (MLJ-wide) show convention. Should we really pursue this goal? Maybe some concepts like loss functions, table transforms deserve a custom display? Maybe maintainers of these modules can provide customized display for these concepts that they maintain to help users grasp what is going on in a way that is more visually appealing? Remember that TableTransforms.jl is project-agnostic and is being used by other frameworks like GeoStats.jl for example. We don't want to impose a project-wide show convention, we just want a self-contained package for tabular transformations that has the "best" display methods possible, where "best" here is defined in terms of end-users, who may not be users of MLJ.jl, GeoStats.jl or any other framework.

ablaom:
@juliohm Thanks for those thoughts.

PR #102, opened 2022-06-17
Restrict Levels to categorical columns

### LINKED ISSUE #95: Consider restricting Levels to columns with categorical scitype
Currently the Levels transform contributed by @Frank-III and refactored by @eliascarv can operate directly on any type of column, including columns with the wrong scitype. I think we could consider a more consistent approach where we ask users to be explicit about the scitypes using the Coerce transform first. Even though this is a breaking change, it will provide more safety in real-world applications of pipelines.

eliascarv:
closes #95

### DIFF
--- a/src/transforms/levels.jl
+++ b/src/transforms/levels.jl
@@ -29,15 +29,6 @@ Levels(; kwargs...) = throw(ArgumentError("Cannot create a Levels object without
 
 isrevertible(transform::Levels) = true
 
-_categorical(x::AbstractVector, l, o) =
-  categorical(x, levels=l, ordered=o), y -> unwrap.(y)
-
-function _categorical(x::CategoricalArray, l, o)
-  xl, xo = levels(x), isordered(x)
-  revfunc = y -> categorical(y, levels=xl, ordered=xo)
-  categorical(x, levels=l, ordered=o), revfunc
-end
-
 function apply(transform::Levels, table)
   cols = Tables.columns(table)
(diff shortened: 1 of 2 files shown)


src/transforms/levels.jl
```diff
@@ -29,28 +29,24 @@ Levels(; kwargs...) = throw(ArgumentError("Cannot create a Levels object without
...
+      assert_categorical(x)
+      
       o = nm ∈ ordered
-      l = levels[findfirst(==(nm), snames)]
-      return _categorical(x, l, o)
+      l = tlevels[findfirst(==(nm), snames)]
+      y = categorical(x, levels=l, ordered=o)
+      xl, xo = levels(x), isordered(x)
+      revfunc = y -> categorical(y, levels=xl, ordered=xo)
+      return y, revfunc
```
juliohm, line 49:
```suggestion
      y, revfunc
```

juliohm (changes requested):
Small adjustment

src/transforms/levels.jl
eliascarv: This `return` is necessary.

juliohm: Oh so maybe we are missing an else branch in the if statement? I reviewed the code quickly and missed that this was an actual case of early termination.

eliascarv: Take a look at the new version of the code, @juliohm.

juliohm (approved):

PR #104, opened 2022-06-22
Add ColSpec to Functional

eliascarv:


src/transforms/functional.jl
```diff
@@ -48,45 +49,57 @@ inverse(::typeof(identity)) = identity
 # fallback to nothing
 inverse(::Any) = nothing
 
-isrevertible(transform::Functional) =
+isrevertible(transform::Functional{S}) where {S} =
```
juliohm, line 52: why do we need to mention S here?

juliohm (changes requested):
Remove unnecessary mentions to type parameters

juliohm (approved):

PR #109, opened 2022-09-14
Add SelectRename

### LINKED ISSUE #106: Add option to rename columns during selection
A nice feature of the DataFrames.jl select function is that it allows in-place renaming:

```julia
select(df, :x => :y, :a => :b)
```

Currently, we have to write a `Select(:x, :a)` and then use `Rename(:x => :y, :a => :b)` on the result. Maybe we could introduce a `SelectRename` transform or generalize the `Select` to accept renaming.

eliascarv:
closes #106

### DIFF
--- a/src/TableTransforms.jl
+++ b/src/TableTransforms.jl
@@ -32,6 +32,7 @@ export
   # built-in
   Select,
   Reject,
+  SelectRename,
   Rename,
   StdNames,
   Sort,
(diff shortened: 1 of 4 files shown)


juliohm:
Thank you @eliascarv , do you think the logic in Select itself could be changed without too much code to accommodate the rename option? Or the types of arguments we accept make it difficult?

juliohm:
Superseded by #110

PR #111, opened 2022-09-15
Refactor Sample

juliohm:
This is a necessary breaking change to avoid the external dependence on StatsBase.jl weights and to make the end-user experience more smooth.

### DIFF
--- a/src/transforms/sample.jl
+++ b/src/transforms/sample.jl
@@ -33,10 +33,18 @@ struct Sample{W,RNG} <: Stateless
   rng::RNG
 end
 
-Sample(size::Int, weights=nothing;
+Sample(size::Int;
+       replace=false, ordered=false,
+       rng=Random.GLOBAL_RNG) =
+  Sample(size, nothing, replace, ordered, rng)
+  
+Sample(size::Int, weights::AbstractWeights;
        replace=false, ordered=false,
        rng=Random.GLOBAL_RNG) =
   Sample(size, weights, replace, ordered, rng)
+  
+Sample(size::Int, weights; kwargs...) =
+  Sample(size, Weights(collect(weights)); kwargs...)
 
 isrevertible(::Type{<:Sample}) = false
 
@@ -52,8 +60,7 @@ function apply(transform::Sample, table)
   newrows = if isnothing(weights)
     sample(rng, rows, size; replace, ordered)
   else
-    ws = Weights(collect(weights))
-    sample(rng, rows, ws, size; replace, ordered)
+    sample(rng, rows, weights, size; replace, ordered)
   end
 
   newtable = newrows |> Tables.materializer(table)
--- a/src/transforms/sample.jl
+++ b/src/transforms/sample.jl
@@ -33,7 +33,7 @@ struct Sample{W,RNG} <: Stateless
   rng::RNG
 end
 
-Sample(size::Int, weights=[];
+Sample(size::Int, weights=nothing;
        replace=false, ordered=false,
        rng=Random.GLOBAL_RNG) =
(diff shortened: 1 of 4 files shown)


src/transforms/sample.jl
```diff
@@ -18,50 +19,41 @@ Sample(1_000, replace=false, ordered=true)
...
   ordered = transform.ordered
+  rng     = transform.rng
 
-  if isnothing(wv)
-    newrows = sample(rng, rows, n; replace, ordered)
+  newrows = if isempty(weights)
+    sample(rng, rows, size; replace, ordered)
   else
-    newrows = sample(rng, rows, wv, n; replace, ordered)
+    ws = Weights(collect(weights))
```
eliascarv, line 55: I think this will limit the implementation, because StatsBase.jl defines other types of weights:
https://juliastats.org/StatsBase.jl/latest/weights/

src/transforms/sample.jl
```diff
@@ -18,50 +19,41 @@ Sample(1_000, replace=false, ordered=true)
...
-
-Sample(wv::AbstractWeights, n::Int; replace=true, ordered=false) =
-  Sample(GLOBAL_RNG, wv, n, replace, ordered)
-
-Sample(n::Int; replace=true, ordered=false) =
-  Sample(GLOBAL_RNG, nothing, n, replace, ordered)
+Sample(size::Int, weights=[];
+       replace=false, ordered=false,
+       rng=Random.GLOBAL_RNG) =
+  Sample(size, weights, replace, ordered, rng)
```
eliascarv, line 39:
```suggestion
Sample(size::Int; replace=false, ordered=false, rng=Random.GLOBAL_RNG) =
  Sample(size, nothing, replace, ordered, rng)
  
Sample(size::Int, weights::AbstractWeights;
       replace=false, ordered=false,
       rng=Random.GLOBAL_RNG) =
  Sample(size, weights, replace, ordered, rng)
  
Sample(size::Int, weights; kwargs...) = Sample(size, Weights(collect(weights)); kwargs...)
```

I think this solution is more generic.

src/transforms/sample.jl
juliohm: Yes, but I think we only ProbabilityWeights (or the generic Weights) in this use case. For end-users it is much more convenient to hide this internal statistical jargon and let them pass any vector of numbers as input, like one of the columns of the table.

src/transforms/sample.jl
juliohm: Our main purpose in this PR is to eliminate this variation of AbstractWeights types that StatsBase.jl introduces. In this use case where we want to sample rows from a table with weights, we just mean probability weights, or weights that are proportional to the inverse frequency.

test/transforms.jl
```diff
@@ -578,15 +578,15 @@
...
     n, c = apply(T, t)
     @test n.a == [3, 7, 8, 2, 2, 6, 2, 6]
     @test n.b == [8, 2, 3, 1, 1, 5, 1, 5]
     @test n.c == [1, 2, 9, 5, 5, 8, 5, 8]
 
     #with weights
-    wv = pweights([0.1, 0.25, 0.15, 0.25, 0.1, 0.15])
-    T = Sample(MersenneTwister(2), wv, 10_000)
+    w = [0.1, 0.25, 0.15, 0.25, 0.1, 0.15]
+    T = Sample(10_000, w, replace=true, rng=MersenneTwister(2))
```
eliascarv, line 589:
```suggestion
    w = pweights([0.1, 0.25, 0.15, 0.25, 0.1, 0.15])
    T = Sample(10_000, w, replace=true, rng=MersenneTwister(2))
```

PR #112, opened 2022-09-17
Add isindexable and indices to public API

juliohm:
Some transforms such as Filter, Sample, and Sort are implemented in terms of row indices. This PR introduces two new trait functions to the public API to facilitate extensions downstream.

@eliascarv I didn't test performance regressions. It would be nice to double check, particularly in the case of Filter if we are doing something suboptimal.


juliohm:
Superseded by #117

PR #115, opened 2022-09-18
Add support for metadata

### LINKED ISSUE #114: Treatment of metadata
This issue tracks our efforts to generalize the API to handle metadata (or special columns). The overall idea consists of renaming all apply, revert, reapply functions to internal functions that operate on the feature table. End users won't feel any difference given that the fallbacks don't change the current behavior.

juliohm:
Fixes #114 

This PR generalizes our API to handle metadata. I needed to introduce two new trait functions `divide` and `attach` to be implemented by table types in order to seamlessly integrate with all our transforms.

### DIFF
--- a/src/TableTransforms.jl
+++ b/src/TableTransforms.jl
@@ -20,6 +20,7 @@ using Random
 import Distributions: ContinuousUnivariateDistribution
 import Distributions: quantile, cdf
 
+include("tabletraits.jl")
 include("assertions.jl")
 include("distributions.jl")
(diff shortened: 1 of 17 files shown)


juliohm:
I realized that the attach trait needs an extra argument to specify the type of the result. Will think about it a bit more.

juliohm:
Nevermind, we can rely on the type of the metadata to specialize the dispatch.

eliascarv (approved):

PR #122, opened 2022-09-19
Fix Parallel transform in the presence of metadata

juliohm:
This is a hotfix for the Parallel transform. I started playing with more complex pipelines and realized that we didn't have this case covered in our tests. In a future PR we should consider introducing a dummy table type with metadata in our test suite.

### DIFF
--- a/src/transforms/parallel.jl
+++ b/src/transforms/parallel.jl
@@ -42,9 +42,9 @@ function apply(p::Parallel, table)
   caches = last.(vals)
 
   # features and metadata
-  divid = divide.(tables)
-  feats = first.(divid)
-  metas = last.(divid)
+  splits = divide.(tables)
+  feats  = first.(splits)
+  metas  = last.(splits)
 
   # table with concatenated features
   newfeat = tablehcat(feats)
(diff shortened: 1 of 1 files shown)


src/transforms/parallel.jl
```diff
@@ -41,8 +41,19 @@ function apply(p::Parallel, table)
   tables = first.(vals)
   caches = last.(vals)
 
-  # table with concatenated columns
-  newtable = tablehcat(tables)
+  # features and metadata
+  divid = divide.(tables)
+  feats = first.(divid)
+  metas = last.(divid)
```
juliohm, line 47:
```suggestion
  splits = divide.(tables)
  feats  = first.(splits)
  metas  = last.(splits)
```

src/transforms/parallel.jl
```diff
@@ -100,8 +119,19 @@ function reapply(p::Parallel, table, cache)
   itr     = zip(p.transforms, caches)
   tables  = tcollect(f(t, c) for (t, c) in itr)
 
-  # table with concatenated columns
-  tablehcat(tables)
+  # features and metadata
+  divid = divide.(tables)
+  feats = first.(divid)
+  metas = last.(divid)
```
juliohm, line 125:
```suggestion
  splits = divide.(tables)
  feats  = first.(splits)
  metas  = last.(splits)
```

eliascarv (approved):

PR #129, opened 2022-09-22
Refactor Colwise

eliascarv:


src/transforms.jl
```diff
@@ -273,10 +283,10 @@ function applyfeat(transform::Colwise, feat, prep)
   newfeat = 𝒯 |> Tables.materializer(feat)
 
   # cache values for each column
-  fcache = last.(vals)
```
juliohm, line 276: Isn't fcache the correct term in this context? We are inside the applyfeat, right?

juliohm (commented):
Amazing work @eliascarv , just left a small comment regarding the rename of a fcache variable. Can you please double check?

src/transforms.jl
eliascarv: In this case, fcache is the tuple of colcaches and snames:
```julia
fcache = (caches, snames)
```

juliohm:
Perfect. Thanks for double checking. Feel free to merge it. :)

Em sex., 23 de set. de 2022 08:07, Elias Carvalho ***@***.***>
escreveu:

PR #139, opened 2022-09-29
Updates for API 0.3

juliohm:
This concludes the migration to `SequentialTransform`. Notice that in order to support combinations of different `Transform` with tables we had to relax the dispatch for `ParallelTableTransform`.


juliohm:
@eliascarv I will go ahead and merge this one to finish the full migration downstream. I understand you are a bit busy today.

PR #140, opened 2022-09-29
Refactor Scale

eliascarv:
This PR makes Scale constructor simpler and avoids type conversion in `apply(transform, table)`.

### DIFF
--- a/src/transforms/scale.jl
+++ b/src/transforms/scale.jl
@@ -35,19 +35,16 @@ Scale(r"[ace]", low=0.3, high=0.7)
 
 * The `low` and `high` values are restricted to the interval [0, 1].
 """
-struct Scale{S<:ColSpec,T<:Real} <: ColwiseFeatureTransform
+struct Scale{S<:ColSpec} <: ColwiseFeatureTransform
   colspec::S
-  low::T
-  high::T
+  low::Float64
+  high::Float64
 
-  function Scale(colspec::S, low::T, high::T) where {S<:ColSpec,T<:Real}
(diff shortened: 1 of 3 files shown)


src/transforms/scale.jl
```diff
@@ -35,20 +35,17 @@ Scale(r"[ace]", low=0.3, high=0.7)
 
 * The `low` and `high` values are restricted to the interval [0, 1].
 """
-struct Scale{S<:ColSpec,T<:Real} <: ColwiseFeatureTransform
+struct Scale{S<:ColSpec} <: ColwiseFeatureTransform
   colspec::S
-  low::T
-  high::T
+  low::Float64
+  high::Float64
```
juliohm, line 41: I think we can still use a type parameter T here to avoid a strict design. The apply function will take care of converting the type to eltype of x anyways.

juliohm (changes requested):
Thank you @eliascarv , review attached.

test/transforms/scale.jl
```diff
@@ -42,7 +42,7 @@
   # columntype does not change
   for FT in (Float16, Float32)
     t = Table(; x=rand(FT, 10))
-    for T in (MinMax(), Scale(low=FT(0), high=FT(0.5)))
+    for T in (MinMax(), Interquartile(), Scale(low=0, high=0.5))
```
juliohm, line 45: Do we need to add a test here to make sure that the result preserves the FT?

juliohm (changes requested):
Review attached

PR #141, opened 2022-09-30
Add ProjectionPursuit (new PR)

DianaPat:


juliohm (changes requested):
We need a more complex multivariate distribution in the tests. The one we have currently is not challenging enough to capture possible bugs in the implementation.

src/transforms/projectionpursuit.jl
```diff
@@ -0,0 +1,199 @@
...
+  # rotate features with orthonormal basis
+  table = Tables.table(Z * Q)
+  
+  # remove structure of first rotated axis
+  newtable, cache = apply(Quantile(1), table)
+  
+  # undo rotation, i.e recover original axis-aligned features
+  Z₊ = Tables.matrix(newtable) * Q'
+  
+  Z₊, Q, cache
```
juliohm, line 146:
```suggestion
  newtable, qcache = apply(Quantile(1), table)
  
  # undo rotation, i.e recover original axis-aligned features
  Z₊ = Tables.matrix(newtable) * Q'
  
  Z₊, (Q, qcache)
```

src/transforms/projectionpursuit.jl
```diff
@@ -0,0 +1,199 @@
...
+  # standard Gaussian quantiles
+  g = gaussquantiles(transform, N, q) 
+
+  iter = 0
+  caches = []
+  while any(g .< I) && iter ≤ transform.maxiter
+    α = alphamax(transform, Z)
+    Z, Q, cache = rmstructure(transform, Z, α)
+    I = pbasis(transform, Z)
+    push!(caches, (Q, cache))
```
juliohm, line 170:
```suggestion
    Z, cache = rmstructure(transform, Z, α)
    I = pbasis(transform, Z)
    push!(caches, cache)
```

src/transforms/projectionpursuit.jl
```diff
@@ -0,0 +1,199 @@
...
+  ttable, tcache = apply(Quantile() → EigenAnalysis(:VDV), table)
+
+  Z = Tables.matrix(ttable)
+  N, q = size(Z)
+  
+  # initialize scores along original axis-aligned features
+  I = pbasis(transform, Z)
+
+  # standard Gaussian quantiles
+  g = gaussquantiles(transform, N, q)
```
juliohm, line 162:
```suggestion
  
  # standard Gaussian quantiles
  g = gaussquantiles(transform, size(Z)...) 
  
  # initialize scores along original axis-aligned features
  I = pbasis(transform, Z)
```

src/transforms/projectionpursuit.jl
```diff
@@ -0,0 +1,198 @@
...
+  
+  # standard Gaussian quantiles
+  g = gaussquantiles(transform, size(Z)...) 
+  
+  # initialize scores along original axis-aligned features
+  I = pbasis(transform, Z)
+
+  iter = 0
+  caches = []
+  while any(g .< I) && iter ≤ transform.maxiter
```
juliohm, line 165:
```suggestion
  while any(I .> g) && iter ≤ transform.maxiter
```

src/transforms/projectionpursuit.jl
```diff
@@ -0,0 +1,198 @@
...
+  Z₊ = Tables.matrix(newtable) * Q'
+  
+  Z₊, (Q, qcache)
+end
+
+function applyfeat(transform::ProjectionPursuit, table, prep) 
+  cols = Tables.columns(table)
+  names = Tables.columnnames(cols)
+
+  ttable, tcache = apply(Quantile() → EigenAnalysis(:VDV), table)
```
juliohm, line 153:
```suggestion
  # preprocess the data to approximately spherical shape
  ptable, pcache = apply(Quantile() → EigenAnalysis(:VDV), table)
```

src/transforms/projectionpursuit.jl
```diff
@@ -0,0 +1,198 @@
...
+
+  Z = Tables.matrix(ttable)
```
juliohm, line 155:
```suggestion
  Z = Tables.matrix(ptable)
```

src/transforms/projectionpursuit.jl
```diff
@@ -0,0 +1,198 @@
...
+    α = alphamax(transform, Z)
+    Z, cache = rmstructure(transform, Z, α)
+    I = pbasis(transform, Z)
+    push!(caches, cache)
+    iter += 1
+  end
+
+  𝒯 = (; zip(names, eachcol(Z))...)
+  newtable = 𝒯 |> Tables.materializer(table)
+  newtable, (caches, tcache)
```
juliohm, line 175:
```suggestion
  newtable, (caches, pcache)
```

src/transforms/projectionpursuit.jl
```diff
@@ -0,0 +1,199 @@
...
+  
+  # remove structure of first rotated axis
+  newtable, qcache = apply(Quantile(1), table)
+  
+  # undo rotation, i.e recover original axis-aligned features
+  Z₊ = Tables.matrix(newtable) * Q'
+  
+  Z₊, (Q, qcache)
+end
+
```
juliohm, line 148:
```suggestion

sphering() = Quantile() → EigenAnalysis(:VDV)

```

src/transforms/projectionpursuit.jl
```diff
@@ -0,0 +1,201 @@
...
+sphering() = Quantile() → EigenAnalysis(:VDV)
+
+function applyfeat(transform::ProjectionPursuit, table, prep) 
+  cols = Tables.columns(table)
+  names = Tables.columnnames(cols)
+
+  # preprocess the data to approximately spherical shape
+  ptable, pcache = apply(Quantile() → EigenAnalysis(:VDV), table)
```
juliohm, line 156:
```suggestion
  ptable, pcache = apply(sphering(), table)
```

src/transforms/projectionpursuit.jl
```diff
@@ -0,0 +1,201 @@
...
+  for (Q, cache) in reverse(caches)
+    # rotate the data 
+    Z = Tables.matrix(t) * Q
+
+    # revert the transform
+    table  = revert(Quantile(1), Tables.table(Z), cache)
+    t = Tables.matrix(table) * Q'
+  end
+  
+  tablerev = revert(Quantile() → EigenAnalysis(:VDV), t, tcache)
```
juliohm, line 196:
```suggestion
  tablerev = revert(sphering(), t, tcache)
```

src/transforms/projectionpursuit.jl
```diff
@@ -0,0 +1,201 @@
...
+  # preprocess the data to approximately spherical shape
+  ptable, pcache = apply(sphering(), table)
+
+  Z = Tables.matrix(ptable)
+  
+  # standard Gaussian quantiles
+  g = gaussquantiles(transform, size(Z)...) 
+  
+  # initialize scores along original axis-aligned features
+  I = pbasis(transform, Z)
```
juliohm, line 164:
```suggestion
  # initialize scores and Gaussian quantiles
  Z = Tables.matrix(ptable)
  I = pbasis(transform, Z)
  g = gaussquantiles(transform, size(Z)...) 
```

src/transforms/projectionpursuit.jl
```diff
@@ -0,0 +1,198 @@
...
+
+  iter = 0
+  caches = []
+  while any(I .> g) && iter ≤ transform.maxiter
+    α = alphamax(transform, Z)
+    Z, cache = rmstructure(transform, Z, α)
+    I = pbasis(transform, Z)
+    push!(caches, cache)
+    iter += 1
+  end
```
juliohm, line 171:
```suggestion
  iter = 0; caches = []
  while any(I .> g) && iter ≤ transform.maxiter
    # choose direction with maximum projection index
    α = alphamax(transform, Z)
    
    # remove non-Gaussian structure
    Z, cache = rmstructure(transform, Z, α)
    
    # update the scores along original axes
    I = pbasis(transform, Z)
    
    # store cache and continue
    push!(caches, cache)
    iter += 1
  end
```

src/transforms/projectionpursuit.jl
```diff
@@ -0,0 +1,204 @@
...
+    # revert the transform
+    table  = revert(Quantile(1), Tables.table(Z), cache)
+    t = Tables.matrix(table) * Q'
+  end
+  
+  tablerev = revert(sphering(), t, tcache)
+
+  Z = Tables.matrix(tablerev)
+  𝒯 = (; zip(names, eachcol(Z))...)
+  newtable = 𝒯 |> Tables.materializer(newtable)
```
juliohm, line 203:
```suggestion
  caches, pcache = fcache

  t = newtable
  for (Q, qcache) in reverse(caches)
    # rotate the data 
    Z = Tables.matrix(t) * Q

    # revert the transform
    table  = revert(Quantile(1), Tables.table(Z), qcache)
    t = Tables.matrix(table) * Q'
  end
  
  tablerev = revert(sphering(), t, pcache)

  Z = Tables.matrix(tablerev)
  𝒯 = (; zip(names, eachcol(Z))...)
  newtable = 𝒯 |> Tables.materializer(newtable)
```

test/transforms/projectionpursuit.jl
```diff
@@ -0,0 +1,50 @@
+@testset "ProjectionPursuit" begin
+    a = rand(100)
+    b = rand(100)
+    c = b + rand(100)
+    t = Table(; a, b, c)
```
juliohm, line 5: Hi @DianaPat can you replace this first example by the example provided in the PairPlots.jl package?

https://github.com/sefffal/PairPlots.jl#examples

Also, can you plot the before and after the transform? Maybe we should add PairPlots.jl as a test dependency to our test/Project.toml? That way we could make use of corner(table) to see the before and after. You can modify the test/Project.toml by activating the test folder and typing `] add PairPlots`

test/transforms/projectionpursuit.jl
```diff
@@ -0,0 +1,50 @@
...
+    @test isapprox(Σ[1,2], 0; atol=1e-6)
+    @test isapprox(Σ[2,1], 0; atol=1e-6)
+    @test isapprox(Σ[2,2], 1; atol=1e-6)
+
+    tₒ = revert(T, n, c)
+  
+    if visualtests
+      p₁ = scatter(t.a, t.b, label="Original")
+      p₂ = scatter(tₒ.a, tₒ.b, label="Reverted")
+      p = plot(p₁, p₂, layout=(1,2))
```
juliohm, line 47: Let's also use the corner plot here to compare the before, after, and then revert.

src/transforms/projectionpursuit.jl
```diff
@@ -0,0 +1,204 @@
...
+  # undo rotation, i.e recover original axis-aligned features
+  Z₊ = Tables.matrix(newtable) * Q'
+  
+  Z₊, (Q, qcache)
+end
+
+sphering() = Quantile() → EigenAnalysis(:VDV)
+
+function applyfeat(transform::ProjectionPursuit, table, prep) 
+  cols = Tables.columns(table)
```
juliohm, line 152:
```suggestion
  # retrieve column names
  cols = Tables.columns(table)
```

src/transforms/projectionpursuit.jl
```diff
@@ -0,0 +1,205 @@
...
+
+  𝒯 = (; zip(names, eachcol(Z))...)
+  newtable = 𝒯 |> Tables.materializer(table)
+  newtable, (caches, pcache)
+end
+
+function revertfeat(::ProjectionPursuit, newtable, fcache)
+  cols = Tables.columns(newtable)
+  names = Tables.columnnames(cols)
+  caches, pcache = fcache
```
juliohm, line 188:
```suggestion
  newtable, (pcache, caches)
end

function revertfeat(::ProjectionPursuit, newtable, fcache)
  # retrieve column names
  cols = Tables.columns(newtable)
  names = Tables.columnnames(cols)
  
  # caches to retrieve transform steps
  pcache, caches = fcache
```

src/transforms/projectionpursuit.jl
```diff
@@ -0,0 +1,208 @@
...
+  for (Q, qcache) in reverse(caches)
+    # rotate the data 
+    Z = Tables.matrix(t) * Q
+
+    # revert the transform
+    table  = revert(Quantile(1), Tables.table(Z), qcache)
+    t = Tables.matrix(table) * Q'
+  end
+  
+  tablerev = revert(sphering(), t, pcache)
```
juliohm, line 203:
```suggestion
  Z = Tables.matrix(newtable)
  for (Q, qcache) in reverse(caches)
    table = revert(Quantile(1), Tables.table(Z * Q), qcache)
    Z = Tables.matrix(table) * Q'
  end
  
  tablerev = revert(sphering(), Tables.table(Z), pcache)
```

src/transforms/projectionpursuit.jl
```diff
@@ -0,0 +1,204 @@
...
+
+  Z = Tables.matrix(newtable)
+  for (Q, qcache) in reverse(caches)
+    table = revert(Quantile(1), Tables.table(Z * Q), qcache)
+    Z = Tables.matrix(table) * Q'
+  end
+  
+  tablerev = revert(sphering(), Tables.table(Z), pcache)
+
+  Z = Tables.matrix(tablerev)
```
juliohm, line 201:
```suggestion
  table = revert(sphering(), Tables.table(Z), pcache)
  Z = Tables.matrix(table)
  
```

src/transforms/projectionpursuit.jl
```diff
@@ -0,0 +1,204 @@
...
+  Q, R = qr([α rand(q,q-1)])
+  while norm(diag(R)) < transform.tol
+    Q, R = qr([α rand(q,q-1)])
+  end
+
+  # rotate features with orthonormal basis
+  table = Tables.table(Z * Q)
+  
+  # remove structure of first rotated axis
+  newtable, qcache = apply(Quantile(1), table)
```
juliohm, line 141:
```suggestion
  # remove structure of first rotated axis
  newtable, qcache = apply(Quantile(1), Tables.table(Z * Q))
```

src/transforms/projectionpursuit.jl
```diff
@@ -0,0 +1,201 @@
...
+  f(α) = -pindex(transform, Z, α)
+  op = optimize(f, α₀)
+  minimizer(op)
+end
+
+function alphamax(transform, Z)
+  α = alphaguess(transform, Z)
+  neldermead(transform, Z, α)  
+end
+
```
juliohm, line 127:
```suggestion

function orthobasis(α, tol)
  q = length(α)
  Q, R = qr([α rand(q,q-1)])
  while norm(diag(R)) < tol
    Q, R = qr([α rand(q,q-1)])
  end  
  Q
end

```

src/transforms/projectionpursuit.jl
```diff
@@ -0,0 +1,210 @@
...
+function rmstructure(transform, Z, α)
+  q = length(α)
+  
+  # find orthonormal basis for rotation
+  Q, R = qr([α rand(q,q-1)])
+  while norm(diag(R)) < transform.tol
+    Q, R = qr([α rand(q,q-1)])
+  end
```
juliohm, line 144:
```suggestion
  # find orthonormal basis for rotation
  Q = orthobasis(α, transform.tol)
```

src/transforms/projectionpursuit.jl
```diff
@@ -0,0 +1,205 @@
...
+
+ProjectionPursuit(;tol=1e-6, maxiter=100, deg=5, perc=.95, n=100) =
+  ProjectionPursuit{typeof(tol)}(tol, maxiter, deg, perc, n)
+
+isrevertible(::Type{<:ProjectionPursuit}) = true
+
+# transforms a row of random variables into a convex combination 
+# of random variables with values in [-1,1] and standard normal distribution
+function rscore(Z, α)
+  ᾱ = (1/norm(α)) .* α
```
juliohm, line 50:
```suggestion
  ᾱ = α ./ norm(α)
```

src/transforms/projectionpursuit.jl
```diff
@@ -0,0 +1,205 @@
...
+  # objective function
+  func(α) = pindex(transform, Z, α)
+  
+  # evaluate objective along axes
+  j = argmax(j -> func(basis(q, j)), 1:q)
+  α = basis(q, j)
+  I = func(α)
+  
+  # evaluate objective along diagonals
+  diag(α, s, e) = (1/√(2+2s*α'*e)) * (α + s*e)
```
juliohm, line 101: @DianaPat can you please double check the norm of these diagonals? I think the normalization constant is there to guarantee that they are all vectors with norm 1?

src/transforms/projectionpursuit.jl
```diff
@@ -0,0 +1,205 @@
...
+  for eᵢ in eachcol(E)
+    d₊ = diag(α, +1, eᵢ)
+    d₋ = diag(α, -1, eᵢ)
+    f₊ = func(d₊)
+    f₋ = α'*eᵢ != 1.0 ? func(d₋) : 0.0
+    f, d = f₊ > f₋ ? (f₊, d₊) : (f₋, d₋)
+    if f > Iₐ
```
juliohm, line 108:
```suggestion
    if f > I
```
I wonder how the code was compiling with this outdated variable name.

src/transforms/projectionpursuit.jl
```diff
@@ -0,0 +1,205 @@
...
+  α = basis(q, j)
+  I = func(α)
+  
+  # evaluate objective along diagonals
+  diag(α, s, e) = (1/√(2+2s*α'*e)) * (α + s*e)
+  for eᵢ in eachcol(E)
+    d₊ = diag(α, +1, eᵢ)
+    d₋ = diag(α, -1, eᵢ)
+    f₊ = func(d₊)
+    f₋ = α'*eᵢ != 1.0 ? func(d₋) : 0.0
```
juliohm, line 106: Many of these dot products α'*eᵢ can be written more efficiently as α \cdot eᵢ

src/transforms/projectionpursuit.jl
juliohm: Also, please use the \cdot operator for dot product, it should be more efficient in general.

DianaPat:
Hi @juliohm 
I already changed to \cdot notation, checked that the diag function really gives a unitary vector and got rid of some outdated notation.
But when I tried to change the plots in the visual test, they didn't get plotted all too well:
projectionpursuit-1 projectionpursuit-2
One of the possible changes would be to modify the gr settings to change the size of the image, but even so the graphics at the diagonal keep coming out as lines (just like in the images above), so for now I just did the test with each corner separately and will continue checking the problem of the diagonals.

juliohm:
> I already changed to \cdot notation, checked that the diag function really gives a unitary vector and got rid of some outdated notation.

That is great @DianaPat , now we can remove the normalization of alpha inside the rscore function, and only normalize it in the neadermead step. I will propose the changes so that you can review and accept.

> But when I tried to change the plots in the visual test, they didn't get plotted all too well:

Let's just plot the result of the corner plot after the transformation in one figure and the result of reverting the transformation in a separate figure. They don't need to be in the same figure. Also, you can pass the option `size=(800,800)` to the specific plot commands to overwrite the default size set in the `gr` command.

src/transforms/projectionpursuit.jl
```diff
@@ -0,0 +1,205 @@
...
+
+isrevertible(::Type{<:ProjectionPursuit}) = true
+
+# transforms a row of random variables into a convex combination 
+# of random variables with values in [-1,1] and standard normal distribution
+function rscore(Z, α)
+  ᾱ = α ./ norm(α)
+  X = Z * ᾱ
+  2 .* cdf.(Normal(), X) .- 1
+end
```
juliohm, line 53:
```suggestion
rscore(Z, α) = 2 .* cdf.(Normal(), Z * α) .- 1
```

src/transforms/projectionpursuit.jl
```diff
@@ -0,0 +1,205 @@
...
+      α = d
+      I = f
+    end
+  end
+  
+  α
+end
+
+function neldermead(transform, Z, α₀)
+  f(α) = -pindex(transform, Z, α)
```
juliohm, line 118:
```suggestion
  f(α) = -pindex(transform, Z, α ./ norm(α))
```
Notice that we only need to normalize α in the Nealder-Mead optimization step. All the previous steps were designed to produce normalized directions already.

test/transforms/projectionpursuit.jl
```diff
@@ -0,0 +1,71 @@
...
+  @test isapprox(Σ[4,2], 0; atol=1e-6)
+  @test isapprox(Σ[4,3], 0; atol=1e-6)
+  @test isapprox(Σ[1,1], 1; atol=1e-6)
+  @test isapprox(Σ[2,2], 1; atol=1e-6)
+  @test isapprox(Σ[3,3], 1; atol=1e-6)
+  @test isapprox(Σ[4,4], 1; atol=1e-6)
+
+  if visualtests
+    p₁ = corner(t, title="Original")
+    p₂ = corner(n, title="Transformed")
```
juliohm, line 40: Please pass the option `size=(..., ...)` to adjust the plots. The current size is inhibiting the visualization of the marginal distributions. Also, notice that the first variable is not Gaussian after the transform so there must be an issue somewhere with the code, or we need more iterations to fix it.

src/transforms/projectionpursuit.jl
```diff
@@ -0,0 +1,201 @@
...
+  ProjectionPursuit{typeof(tol)}(tol, maxiter, deg, perc, n)
+
+isrevertible(::Type{<:ProjectionPursuit}) = true
+
+# transforms a row of random variables into a convex combination 
+# of random variables with values in [-1,1] and standard normal distribution
+rscore(Z, α) = 2 .* cdf.(Normal(), Z * α) .- 1
+
+# projection index of sample along a given direction
+function pindex(transform, Z, α)
```
juliohm, line 52: let's rename pindex to projecindex

src/transforms/projectionpursuit.jl
```diff
@@ -0,0 +1,201 @@
...
+    end
+  end
+  I
+end
+
+# j-th element of the canonical basis in ℝᵈ
+basis(d, j) = 1:d .== j
+
+# index for all vectors in the canonical basis
+function pbasis(transform, Z)
```
juliohm, line 71: let's rename pbasis to projecbasis

test/transforms/projectionpursuit.jl
juliohm: Hi @DianaPat did you have a chance to take a look into this final adjustment? We just need to make sure that the two datasets are plotted correctly. We can take care of the remaining adjustments moving forward.

DianaPat: Hi @juliohm, I have not worked much of it this last few days, but for what I have checked, the marginal distributions don't plot correctly even if I change the size. The PairPlot options don't work either. I think it's a matter of the gr definition, as this only happen when I set it up.

juliohm: I think you need to increase the plot size a bit more so that the marginals don't get shrunk. It is a Plots.jl issue. Can you try with a larger size to see if it works? Something large like 1000x1000 for example just to see if it works, and then you can decrease to a minimum size that works.

DianaPat: It doesn't work.

juliohm:
Hi @DianaPat can you please double check that this GitHub option is enabled on this PR?

https://docs.github.com/en/pull-requests/collaborating-with-pull-requests/working-with-forks/allowing-changes-to-a-pull-request-branch-created-from-a-fork

I fixed the problem locally and would like to push the changes to the PR.

DianaPat:
It is enabled

juliohm:
Hi @DianaPat it is saying that permission is denied. Can you incorporate the fixes in your local copy and regenerate the figures?

1. Remove the option `aspectratio=:equal` from the runtests.jl file and place it in every existing plot command that existed before in the test files. Basically, scale.jl, zscore.jl, eigenanalysis.jl, if I recall correctly.
2. Delete all png files in the test/data directory and rerun the tests to generate new figures.
3. You should see all corner plots correctly plotted now. The `aspectratio` option was the issue.

After you make the changes you will see that the projection pursuit result is not looking quite right yet. The final distribution is not spherical visually indicating that either the number of iterations wasn't enough or that some other bug is still hidden somewhere.

DianaPat:
Yes, I'm already looking for this bug and other ones I've found. I'll update as soon as I find something!

DianaPat:
Hi @juliohm, I think the error is fixed. 

In fact the function was not comparing the data to a gaussian up to `perc` percentile, but with `1-perc`. There was also an issue with just taking a higher percentile because the optim function raised an error. That part was just fixed with the float at line 93.
But then the tests stopped passing (at this point I'm wondering how did they pass before), so I relaxed the conditions.

At this point I think maybe it would be better to change all occurrences of `perc` to `1-perc` in the code so we can keep the previous value, as it is more intuitive.

juliohm:
Hi @DianaPat thank you for checking it. Can you please move the `float` call to within the definition of the `basis` function? This function should return a floating point vector always to avoid type instability.

Regarding the percentile option, you mean that it is more intuitive to use the value .99? It would be great if you could make the appropriate changes. After that, it would be interesting to print the number of iterations of structure removal that are happening with the test cases. It seems that the structure is not quite Gaussian still? The point cloud has some borders that differ visually from a multivariate Gaussian.

DianaPat:
I already made some experiments. Without any modification of parameters it runs the structure removal cycle 23 times. By changing the parameters a little, it reached 101 iterations; the problem is that it took aprox. 300 seg, so the tests would take more time.

I saved the figures to do a comparison of improvement:

test2 test

juliohm:
I am finding it strange that the the points near the "tails" of the resulting distribution are somewhat spread. If you plot the ideal standard Gaussian you will be able to see the difference. Do you know how to use the VSCode debugger? You could add a plot command inside the structure removal loop and then add a breakpoint to see the plot at each iteration. Another thing you can do is check the optimal alphas as we did in class to see if everything is working as expected. Finally, a good diagnostic would be to plot the q-q plot of the transformed columns against the standard Gaussian to see if they pass through the identity line.

DianaPat:
You are right!
The tails diverge from a gaussian outside the interval [-3,3]. I double checked the removal structure function but it's working as expected. 
I went back to the bibliography (Friedman pg 255) and found this:
Tails_under
When I tried increasing the `deg` argument, there was a slight improvement, but it came at the cost of time.

juliohm:
That is a very good catch. Can you comment on the time increase? Is it 2x
slower? How the plots look like with higher degrees? Can we find a good
compromise by default?

Em dom., 16 de out. de 2022 19:46, DianaPat ***@***.***>
escreveu:

DianaPat:
The time increase is ~3.5x: 
With `deg=5` it took around 16 sec to obtain the answer (image on the left), and with `deg=30` it took 352 sec (image on the right)
qq1 qq30

juliohm:
Thank you @DianaPat for these careful checks. I agree with you that the time cost does not pay off, and we just need to be aware of this behavior near the tails.

I will try to review the PR one more time during the day. It would be important to plot the transformed and reversed data for both data sets if that is not already done.

juliohm:
That is really awesome @DianaPat. I think we finally made it! 🎉 Congrats!

One additional change we could implement to avoid depending on the heavy Optim.jl package is replace it with the NelderMead.jl package, which only implements the algorithm are currently using: https://github.com/jwscook/NelderMead.jl

You can learn more about the algorithm in this lecture: https://youtu.be/vOYlVvT3W80 It is beautiful and simple idea.

In theory, we would just need to replace the `Optim.optimize` function by the `NelderMead.optimise` function and rerun the tests. If you feel that you have some time to try this out, please go ahead. Otherwise, I invite you to review the original PR you submitted to compare how much you improved your coding skills in the process 👏🏽 

These picky reviews pay off in the long run, and I am really happy you went all the way to the end! 🥇

DianaPat:
Thank you, I learned a lot from this experience!

On the other hand, I tried to implement the change of package, but it seems that the autor have not registered the package, so I couldn't do it. For now will leave it as is while I look for other alternatives.

juliohm:
I've opened an issue requesting registration: https://github.com/jwscook/NelderMead.jl/issues/5

juliohm:
@DianaPat NelderMead.jl is now registered 👌🏽 would you like to give it a try here to see if the results are the same?

PR #146, opened 2022-10-09
Fix revert for ParallelTableTrasform

juliohm:
Everything works as expected now.

### DIFF
--- a/test/metadata.jl
+++ b/test/metadata.jl
@@ -4,7 +4,7 @@
     b = rand(10)
     c = rand(10)
     d = rand(10)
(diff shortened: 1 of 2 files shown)


test/metadata.jl
```diff
@@ -4,7 +4,7 @@
     b = rand(10)
     c = rand(10)
     d = rand(10)
-    t = Table(; a, b, c, d)
+    t = TypedTables.Table(; a, b, c, d)
```
juliohm, line 7:
```suggestion
    t = Table(; a, b, c, d)
```

eliascarv (approved):

PR #155, opened 2022-11-23
Swap out Optim.jl NelderMead.jl

jwscook:
This PR removes Optim.jl as a dependency and replaces it with [NelderMead.jl](https://github.com/jwscook/NelderMead.jl) given the context of https://github.com/jwscook/NelderMead.jl/issues/7.

A point worth noting is that the size of the simplex is taken to be the standard deviation of the starting position (or 1 in the case that the standard deviation is zero). There may be a better simplex size given the range of sensible values of $\alpha_0$; please advise.

Closes https://github.com/jwscook/NelderMead.jl/issues/7.


src/TableTransforms.jl
```diff
@@ -16,7 +16,7 @@ using PrettyTables
 using AbstractTrees
 using CategoricalArrays
 using Random
-using Optim: optimize, minimizer
+using NelderMead
```
juliohm, line 19:
```suggestion
using NelderMead: optimise
```

src/transforms/projectionpursuit.jl
```diff
@@ -111,8 +111,10 @@ end
 
 function neldermead(transform, Z, α₀)
   f(α) = -pindex(transform, Z, α ./ norm(α))
-  op = optimize(f, α₀)
-  minimizer(op)
+  l = std(α₀)
+  (minimiser, _) = NelderMead.optimise(f, α₀, iszero(l) ? one(l) : l,
+                                       xtol_rel=10eps())
+  minimiser
```
juliohm, line 117:
```suggestion
  res = optimise(f, α₀, iszero(l) ? one(l) : l, xtol_rel=10eps())
  first(res)
```

test/Project.toml
```diff
@@ -11,6 +11,7 @@ ReferenceTests = "324d217c-45ce-50fc-942e-d289b448e8cf"
 ScientificTypes = "321657f4-b219-11e9-178b-2701a2544e81"
 Statistics = "10745b16-79ce-11e8-11f9-7d13ad32a3b2"
 StatsBase = "2913bbd2-ae8a-5f71-8c99-4fb6c76f3a91"
+TableTransforms = "0d432bfd-3ee1-4ac1-886a-39f05cc69a3e"
```
juliohm, line 14:
```suggestion
```

juliohm (changes requested):
Thank you @jwscook for the PR ❤️ Regarding the initial guess, it is guaranteed to have norm 1. Do you have suggestions for the simplex size in this case?

test/Project.toml
jwscook: Woops! Good catch.

jwscook:
> Thank you @jwscook for the PR ❤️ Regarding the initial guess, it is guaranteed to have norm 1. Do you have suggestions for the simplex size in this case?

I'm not entirely sure what the right size should be, but after some trial and error a value of 1/2 seems to work.

src/transforms/projectionpursuit.jl
```diff
@@ -111,8 +111,8 @@ end
 
 function neldermead(transform, Z, α₀)
   f(α) = -pindex(transform, Z, α ./ norm(α))
-  op = optimize(f, α₀)
-  minimizer(op)
+  res = optimise(f, α₀, 1/2, xtol_rel=10eps())
+  first(res)
```
juliohm, line 115:
```suggestion
  optimise(f, α₀, 1/2, xtol_rel=10eps()) |> first
```

juliohm (approved):

juliohm:
That is lovely @jwscook. I will just wait for the tests to pass once again and will then merge it. Did you consider moving NelderMead.jl to JuliaOptim or some other organization with multiple maintainers? It should get more visibility there. :)

PR #160, opened 2022-12-09
fixed AssertionError

vickydeka:
@juliohm  I fixed the Assertion Error and it is working fine now, then again was getting ```ERROR: MethodError: no method matching getindex(::NamedTuple{(:a, :b), Tuple{Vector{Int64}, Vector{Float64}}}, ::Colon, ::Vector{Symbol})``` for 
```jl
(a=rand(Int,3), b=rand(3)) |> ZScore(:b)
```
I guess the ZScore function expects the input data to be a tabular data structure, like DataFrame or a Table, but the input data in this case is a NamedTuple. Converting into dataframe, I am getting correct output:

```jl
julia> DataFrame((a=rand(Int,3), b=rand(3))) |> ZScore(:b)
3×2 DataFrame
 Row │ a                    b
     │ Int64                Float64
─────┼────────────────────────────────
   1 │ 4442619791281385888  -1.12529
   2 │ 7630774422502565167   0.786904
   3 │ 1978852580104395244   0.338386
```

### DIFF
--- a/src/transforms.jl
+++ b/src/transforms.jl
@@ -171,14 +171,15 @@ reapply(transform::StatelessFeatureTransform, table, cache) =
 
 function applyfeat(transform::ColwiseFeatureTransform, feat, prep)
   # retrieve column names and values
-  cols   = Tables.columns(feat) 
+  cols   = Tables.columns(feat)
   names  = Tables.columnnames(cols)
-  snames = choose(transform.colspec, names) 
+  snames = choose(transform.colspec, names)
+  scols = feat |> Select(snames)
 
-  # basic checks 
-  for assertion in assertions(transform) 
+  # basic checks
+  for assertion in assertions(transform)
     # Perform the assertion on the selected columns of the input table
-    assertion(feat |> Select(snames)) 
+    assertion(scols)
   end
   
   # function to transform a single column
--- a/src/transforms.jl
+++ b/src/transforms.jl
@@ -178,7 +178,7 @@ function applyfeat(transform::ColwiseFeatureTransform, feat, prep)
   # basic checks 
   for assertion in assertions(transform) 
     # Perform the assertion on the selected columns of the input table
-    assertion(feat[:, snames]) 
+    assertion(feat |> Select(snames)) 
   end
   
   # function to transform a single column
--- a/src/transforms.jl
+++ b/src/transforms.jl
@@ -174,12 +174,12 @@ function applyfeat(transform::ColwiseFeatureTransform, feat, prep)
   cols   = Tables.columns(feat)
   names  = Tables.columnnames(cols)
   snames = choose(transform.colspec, names)
-  scols = feat |> Select(snames)
+  sfeat  = feat |> Select(snames)
 
   # basic checks
   for assertion in assertions(transform)
     # Perform the assertion on the selected columns of the input table
-    assertion(scols)
+    assertion(sfeat)
   end
   
   # function to transform a single column
--- a/src/transforms.jl
+++ b/src/transforms.jl
@@ -170,16 +170,17 @@ reapply(transform::StatelessFeatureTransform, table, cache) =
 # ------------------
 
 function applyfeat(transform::ColwiseFeatureTransform, feat, prep)
-  # basic checks
-  for assertion in assertions(transform)
-    assertion(feat)
-  end
-
   # retrieve column names and values
-  cols   = Tables.columns(feat)
+  cols   = Tables.columns(feat) 
   names  = Tables.columnnames(cols)
-  snames = choose(transform.colspec, names)
+  snames = choose(transform.colspec, names) 
 
+  # basic checks 
+  for assertion in assertions(transform) 
+    # Perform the assertion on the selected columns of the input table
+    assertion(feat[:, snames]) 
+  end
+  
   # function to transform a single column
   function colfunc(n)
     x = Tables.getcolumn(cols, n)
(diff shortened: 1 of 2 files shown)


src/transforms.jl
```diff
@@ -170,16 +170,17 @@ reapply(transform::StatelessFeatureTransform, table, cache) =
...
 
 function applyfeat(transform::ColwiseFeatureTransform, feat, prep)
-  # basic checks
-  for assertion in assertions(transform)
-    assertion(feat)
-  end
-
   # retrieve column names and values
-  cols   = Tables.columns(feat)
+  cols   = Tables.columns(feat)
```
juliohm, line 174:
```suggestion
  cols   = Tables.columns(feat)
```

src/transforms.jl
```diff
@@ -170,16 +170,17 @@ reapply(transform::StatelessFeatureTransform, table, cache) =
...
-
   # retrieve column names and values
   cols   = Tables.columns(feat)
   names  = Tables.columnnames(cols)
   snames = choose(transform.colspec, names)
 
+  # basic checks 
+  for assertion in assertions(transform) 
+    # Perform the assertion on the selected columns of the input table
+    assertion(feat[:, snames])
```
juliohm, line 181: This syntax is specific to DataFrame. You can replace it with `feat |> Select(snames)` to get a new table with only the selected columns.

vickydeka: Yeah, Now it working fine, Thanks!
```jl
julia> (a=rand(Int,3), b=rand(3)) |> ZScore(:b)
(a = [983918827768240371, 3725987254810195682, -5701988653384801394], b = [0.2585094601557376, 0.8453629612542206, -1.1038724214099582])
```

src/transforms.jl
juliohm, line 174: You are adding whitespace accidentally. I fixed it and you added them back.

juliohm (changes requested):
Please add the example that was failing to the test suite so that the issue is not reintroduced.

src/transforms.jl
```diff
@@ -170,16 +170,17 @@ reapply(transform::StatelessFeatureTransform, table, cache) =
...
-  for assertion in assertions(transform)
-    assertion(feat)
-  end
-
   # retrieve column names and values
   cols   = Tables.columns(feat)
   names  = Tables.columnnames(cols)
   snames = choose(transform.colspec, names)
-
+  selected_cols = feat |> Select(snames)
```
juliohm, line 177: this is not following the code style of the surrounding code. can you please avoid underscores and choose a short and yet descriptive name to represent the selection? also this is missing a new line before the next comment

vickydeka: Okay, maybe I can keep that as scols

juliohm: the result is a table of selected features so maybe sfeat would be more appropriate

vickydeka: Okay, now it is done.

juliohm:
Still missing tests as suggested in the review.

src/transforms.jl
```diff
@@ -170,16 +170,18 @@ reapply(transform::StatelessFeatureTransform, table, cache) =
...
-
   # retrieve column names and values
   cols   = Tables.columns(feat)
   names  = Tables.columnnames(cols)
   snames = choose(transform.colspec, names)
+  sfeat  = feat |> Select(snames)
 
+  # basic checks
+  for assertion in assertions(transform)
+    # Perform the assertion on the selected columns of the input table
```
juliohm, line 181:
```suggestion
```

vickydeka:
> Still missing tests as suggested in the review.

I am not exactly sure, how to add that.

juliohm:
You can copy the example that was failing in our test/transforms/zscore.jl file and add check the expected result. You can find examples in the file. After you add these, GitHub will run them automatically to make sure that the tests are passing in all platforms.

vickydeka:
> and add check the expected res

```jl
a=rand(Int,3) 
b=rand(3)
t = Table(; a, b)
T = ZScore(:b)
n, c = apply(T, t)
@test isapprox(mean(n.b), 0; atol=1e-6)
@test isapprox(std(n.b), 1; atol=1e-6)
tₒ = revert(T, n, c)
@test Tables.matrix(t) ≈ Tables.matrix(tₒ)
```
  Will in this way work?

juliohm:
That should work 👍🏻

Em sáb., 10 de dez. de 2022 02:42, vickydeka ***@***.***>
escreveu:

vickydeka:
> Please add the example that was failing to the test suite so that the issue is not reintroduced.

It is now added.

test/transforms/zscore.jl
```diff
@@ -86,4 +86,14 @@
   @test isapprox(σ[2], 1; atol=1e-6)
   tₒ = revert(T, n, c)
   @test Tables.matrix(t) ≈ Tables.matrix(tₒ)
+  
+  a=rand(Int,3)
+  b=rand(3)
```
eliascarv, line 91:
```suggestion
  a = rand(Int, 10)
  b = rand(10)
```

eliascarv, line 91:
```suggestion
  a = rand(Int, 10)
  b = rand(10)
```
Slightly increase the number of elements of vectors `a` and `b`.

eliascarv (approved):

juliohm (approved):

juliohm:
Thank you @vickydeka for the contribution 💯 The tests are failing for reasons that are not related to your changes.

PR #161, opened 2022-12-20
feat: adds optional `categorical` paramater as option to onehot trans…

### LINKED ISSUE #158: OneHot should probably produce categorical arrays
Currently the resulting columns are just Vector{Bool}, maybe they should be categorical arrays.

rajatrc1705:
…form

fixes #158 

### Default Behavior (categorical is true by default)

image

### Behavior When Categorical is set to false

image

### DIFF
--- a/src/transforms/onehot.jl
+++ b/src/transforms/onehot.jl
@@ -21,7 +21,7 @@ OneHot("a")
 struct OneHot{S<:ColSpec} <: StatelessFeatureTransform
   colspec::S
   categorical::Bool
-  function OneHot(col::Col, categorical::Bool=true)
+  function OneHot(col, categorical)
     cs = colspec([col])
     new{typeof(cs)}(cs, categorical)
   end
--- a/src/transforms/onehot.jl
+++ b/src/transforms/onehot.jl
@@ -55,7 +55,7 @@ function applyfeat(transform::OneHot, feat, prep)
   newnms, newcols = first.(onehot), last.(onehot)
 
   # convert to categorical arrays if necessary
-  newcols = transform.categorical ? categorical.(newcols) : newcols
+  newcols = transform.categorical ? categorical.(newcols, levels=[true, false]) : newcols
   
   splice!(names, ind, newnms)
   splice!(columns, ind, newcols)
@@ -74,7 +74,7 @@ function revertfeat(::OneHot, newfeat, fcache)
   
   oname, inds, levels, ordered = fcache
   x = map(zip(columns[inds]...)) do row
-    levels[findfirst(row)]
+    levels[findfirst(==(true), row)]
   end
 
   ocolumn = categorical(x; levels, ordered)
--- a/src/transforms/onehot.jl
+++ b/src/transforms/onehot.jl
@@ -55,7 +55,7 @@ function applyfeat(transform::OneHot, feat, prep)
   newnms, newcols = first.(onehot), last.(onehot)
 
   # convert to categorical arrays if necessary
-  newcols = transform.categorical ? categorical.(newcols, levels=[true, false]) : newcols
+  newcols = transform.categorical ? categorical.(newcols, levels=[false, true]) : newcols
   
   splice!(names, ind, newnms)
   splice!(columns, ind, newcols)
--- a/src/transforms/onehot.jl
+++ b/src/transforms/onehot.jl
@@ -16,16 +16,19 @@ columns to categorical arrays as opposed to boolean vectors.
 OneHot(1)
 OneHot(:a)
 OneHot("a")
+OneHot("a", categorical=false)
 ```
 """
-struct OneHot{S<:ColSpec} <: StatelessFeatureTransform
+struct OneHot{S<:ColSpec} <: StatelessFeatureTransform 
   colspec::S
   categorical::Bool
   function OneHot(col, categorical)
     cs = colspec([col])
-    new{typeof(cs)}(cs, categorical)
+    new{typeof(cs)}(cs, categorical) 
   end
 end
+
+OneHot(col; categorical=true) = OneHot(col, categorical)
 
 isrevertible(::Type{<:OneHot}) = true
 
--- a/src/transforms/onehot.jl
+++ b/src/transforms/onehot.jl
@@ -3,7 +3,7 @@
 # ------------------------------------------------------------------
 
 """
-    OneHot(col, categorical)
+    OneHot(col; categorical=true)
 
 Transforms categorical column `col` into one-hot columns of levels
 returned by the `levels` function of CategoricalArrays.jl.
--- a/src/transforms/onehot.jl
+++ b/src/transforms/onehot.jl
@@ -19,12 +19,12 @@ OneHot("a")
 OneHot("a", categorical=false)
 ```
 """
-struct OneHot{S<:ColSpec} <: StatelessFeatureTransform 
+struct OneHot{S<:ColSpec} <: StatelessFeatureTransform
   colspec::S
   categorical::Bool
   function OneHot(col, categorical)
     cs = colspec([col])
-    new{typeof(cs)}(cs, categorical) 
+    new{typeof(cs)}(cs, categorical)
   end
 end
 
--- a/src/transforms/onehot.jl
+++ b/src/transforms/onehot.jl
@@ -7,9 +7,9 @@
 
 Transforms categorical column `col` into one-hot columns of levels
 returned by the `levels` function of CategoricalArrays.jl.
-The `categorical` parameter is a bool type parameter, if set to true, 
-the one-hot columns will be categorical vectors, if set to false, 
-the one-hot columns will be bit vectors.
+The `categorical` parameter is a bool type parameter; 
+if set to true the one-hot columns will be categorical vectors, 
+if set to false the one-hot columns will be bit vectors.
 
 # Examples
 
--- a/src/transforms/onehot.jl
+++ b/src/transforms/onehot.jl
@@ -50,8 +50,8 @@ function applyfeat(transform::OneHot, feat, prep)
   end
 
   newnms, newcols = first.(onehot), last.(onehot)
-  if transform.categorical == true
-    newcols = [categorical(newcol; levels=nothing, ordered=false, compress=false) for newcol in newcols]
+  if transform.categorical
+    newcols = categorical.(newcols)
   end
   splice!(names, ind, newnms)
   splice!(columns, ind, newcols)
--- a/src/transforms/onehot.jl
+++ b/src/transforms/onehot.jl
@@ -7,13 +7,10 @@
 
 Transforms categorical column `col` into one-hot columns of levels
 returned by the `levels` function of CategoricalArrays.jl.
-The `categorical` parameter is a bool type parameter, with `true` as
-default value.
-When `categorical` parameter is:
-  (i)  true:  the one-hot columns are of categorical type with values 
-            of either `true` or `false`
-  (ii) false: the one-hot columns are of bit type with values 
-            of either `1` or `0`
+The `categorical` parameter is a bool type parameter, if set to true, 
+the one-hot columns will be categorical vectors, if set to false, 
+the one-hot columns will be bit vectors.
+
 # Examples
 
 ```julia
@@ -78,6 +75,7 @@ function revertfeat(::OneHot, newfeat, fcache)
   end
 
   ocolumn = categorical(x; levels, ordered)
+
   splice!(names, inds, [oname])
   splice!(columns, inds, [ocolumn])
 
--- a/src/transforms/onehot.jl
+++ b/src/transforms/onehot.jl
@@ -50,7 +50,7 @@ function applyfeat(transform::OneHot, feat, prep)
   end
 
   newnms, newcols = first.(onehot), last.(onehot)
-  if transform.category
+  if transform.categorical
     newcols = [CategoricalArray(new_column, ordered=true) for new_column in newcols]
   end
   splice!(names, ind, newnms)
--- a/src/transforms/onehot.jl
+++ b/src/transforms/onehot.jl
@@ -4,12 +4,16 @@
 
 """
     OneHot(col, categorical)
-    
+
 Transforms categorical column `col` into one-hot columns of levels
 returned by the `levels` function of CategoricalArrays.jl.
-Returns CategoryType columns by default. The `categorical` parameter
-can be set to false to get Bool columns after transformation.
-
+The `categorical` parameter is a bool type parameter, with `true` as
+default value.
+When `categorical` parameter is:
+  (i)  true:  the one-hot columns are of categorical type with values 
+            of either `true` or `false`
+  (ii) false: the one-hot columns are of bit type with values 
+            of either `1` or `0`
 # Examples
 
 ```julia
@@ -18,12 +22,12 @@ OneHot(:a)
 OneHot("a")
 ```
 """
-struct OneHot{S<:ColSpec, T<:Bool} <: StatelessFeatureTransform
+struct OneHot{S<:ColSpec} <: StatelessFeatureTransform
   colspec::S
-  categorical::T
+  categorical::Bool
   function OneHot(col::Col, categorical::Bool=true)
     cs = colspec([col])
-    new{typeof(cs), typeof(categorical)}(cs, categorical)
+    new{typeof(cs)}(cs, categorical)
   end
 end
 
@@ -50,8 +54,8 @@ function applyfeat(transform::OneHot, feat, prep)
   end
 
   newnms, newcols = first.(onehot), last.(onehot)
-  if transform.categorical
-    newcols = [CategoricalArray(new_column, ordered=true) for new_column in newcols]
+  if transform.categorical == true
+    newcols = [categorical(newcol; levels=nothing, ordered=false, compress=false) for newcol in newcols]
   end
   splice!(names, ind, newnms)
   splice!(columns, ind, newcols)
@@ -74,7 +78,6 @@ function revertfeat(::OneHot, newfeat, fcache)
   end
 
   ocolumn = categorical(x; levels, ordered)
-
   splice!(names, inds, [oname])
   splice!(columns, inds, [ocolumn])
 
--- a/src/transforms/onehot.jl
+++ b/src/transforms/onehot.jl
@@ -7,9 +7,8 @@
 
 Transforms categorical column `col` into one-hot columns of levels
 returned by the `levels` function of CategoricalArrays.jl.
-The `categorical` parameter is a bool type parameter; 
-if set to true the one-hot columns will be categorical vectors, 
-if set to false the one-hot columns will be bit vectors.
+The `categorical` option can be used to convert resulting
+columns to categorical arrays as opposed to boolean vectors.
 
 # Examples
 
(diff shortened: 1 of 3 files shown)


src/transforms/onehot.jl
```diff
@@ -47,7 +50,9 @@ function applyfeat(transform::OneHot, feat, prep)
   end
 
   newnms, newcols = first.(onehot), last.(onehot)
-
+  if transform.categorical
+    newcols = [CategoricalArray(new_column, ordered=true) for new_column in newcols]
```
juliohm, line 54: We don't use underscore in our code style. Also, the function `categorical(col)` should be used instead of the array constructor.

juliohm (changes requested):
Review attached.

Missing tests.

src/transforms/onehot.jl
rajatrc1705: I will work on these reviews and commit the recommened changes 👍
Regarding the ordered set to "true", I found that it is enabling the comparision of elements using operators like "<", ">",
if ordered is set to "false", comparing elements of the array throws an error, which might have been confusing for the user.

src/transforms/onehot.jl
```diff
@@ -16,11 +18,12 @@ OneHot(:a)
 OneHot("a")
 ```
 """
-struct OneHot{S<:ColSpec} <: StatelessFeatureTransform
+struct OneHot{S<:ColSpec, T<:Bool} <: StatelessFeatureTransform
```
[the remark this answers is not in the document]
eliascarv: The type parameter is not needed here.
Suggestion:
```julia
struct OneHot{S<:ColSpec} <: StatelessFeatureTransform 
  colspec::S
  categorical::Bool
  function OneHot(col::Col; categorical=true)
    cs = colspec([col]) 
    new{typeof(cs)}(cs, categorical) 
  end 
end
```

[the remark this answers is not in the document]
rajatrc1705: Oh great, the changes I made were throwing some error and that was the workaround I came up with. Thank you for the code suggestion 👍

src/transforms/onehot.jl
juliohm: > Regarding the ordered set to "true", I found that it is enabling the comparision of elements using operators like "<", ">",

I think we don't want to enable these relations with binary variables. A binary variable is a like a yes/no variable with no particular ordering. It is often used for masking or grouping two different subsets of a table.

rajatrc1705: Sure thing 👍
I will keep the output unordered (ordered=false)

juliohm: Thank you @rajatrc1705 , looking forward to review it.

juliohm:
@rajatrc1705 did you have a chance to look into this?

rajatrc1705:
@juliohm yes I am on it, currently having some trouble with the `revertfeat` function (failing the tests for onehot.jl), so troubleshooting that part. Will push the changes soon. Apologies if this is taking longer than usual! Will do better for the next contri.

juliohm:
No worries @rajatrc1705 ! Just wanted to double check you are still with us :)

rajatrc1705:
I am still working on the `revertfeat` function, its taking some time, so meanwhile I have pushed the other changes that I have made.

src/transforms/onehot.jl
```diff
@@ -18,9 +24,10 @@ OneHot("a")
 """
 struct OneHot{S<:ColSpec} <: StatelessFeatureTransform
   colspec::S
-  function OneHot(col::Col)
+  categorical::Bool
+  function OneHot(col::Col, categorical::Bool=true)
```
juliohm, line 28: Inner constructors don't take default values usually

src/transforms/onehot.jl
```diff
@@ -3,11 +3,17 @@
...
 Transforms categorical column `col` into one-hot columns of levels
 returned by the `levels` function of CategoricalArrays.jl.
-
+The `categorical` parameter is a bool type parameter, with `true` as
+default value.
+When `categorical` parameter is:
+  (i)  true:  the one-hot columns are of categorical type with values 
+            of either `true` or `false`
+  (ii) false: the one-hot columns are of bit type with values 
+            of either `1` or `0`
```
juliohm, line 16: Be more succinct. People are often in a hurry to learn the basic usage. The less text the better. Just say something like "if categorical is set to `true` the new columns will be categorical arrays".

src/transforms/onehot.jl
```diff
@@ -47,7 +54,9 @@ function applyfeat(transform::OneHot, feat, prep)
   end
 
   newnms, newcols = first.(onehot), last.(onehot)
-
+  if transform.categorical == true
+    newcols = [categorical(newcol; levels=nothing, ordered=false, compress=false) for newcol in newcols]
```
juliohm, line 58: Do we need to pass all these keywords explicitly? Isn't a simple `categorical(newcol)` enough?

src/transforms/onehot.jl
```diff
@@ -69,7 +78,6 @@ function revertfeat(::OneHot, newfeat, fcache)
   end
 
   ocolumn = categorical(x; levels, ordered)
-
```
juliohm, line 72: You are changing the whitespace of the code, deleting break lines. Please avoid that.

test/shows.jl
```diff
@@ -200,13 +200,14 @@
 
     # compact mode
     iostr = sprint(show, T)
-    @test iostr == "OneHot([:a])"
+    @test iostr == "OneHot([:a], true)"
```
juliohm, line 203: categorical must be a keyword argument, not a positional argument.

test/transforms/onehot.jl
```diff
@@ -4,28 +4,31 @@
...
 
-  T = OneHot(1)
+  T = OneHot(1, false)
   n, c = apply(T, t)
   @test Tables.columnnames(n) == (:a_false, :a_true, :b, :c)
-  @test n.a_false == Bool[1, 0, 0, 1, 0, 0]
-  @test n.a_true  == Bool[0, 1, 1, 0, 1, 1]
+  @test n.a_false == categorical(Bool[1, 0, 0, 1, 0, 0])
+  @test n.a_true  == categorical(Bool[0, 1, 1, 0, 1, 1])
+  @test typeof(n.a_false) != typeof(categorical(Bool[1, 0, 0, 1, 0, 0]))
```
juliohm, line 12: We don't need to change the existing tests. We just need to pass the correct option to preserve old results.

juliohm (changes requested):
Review attached.

test/transforms/onehot.jl
rajatrc1705: The existing tests aren't able to distinguish between the categorical columns and bool columns. So I added some modifications so that the tests are able to distinguish.

src/transforms/onehot.jl
rajatrc1705: Okay 👍I will maintain the break lines.

src/transforms/onehot.jl
rajatrc1705: Yes `categorical(newcol)` is enough, but I wanted to explicitly mention the parameters for the first time to make it easier for anyone going through the code.

src/transforms/onehot.jl
rajatrc1705: Sure 👍Will try to be more concise.

rajatrc1705:
Also.. do beginners usually struggle so much with their first PRs 😅, it seems I am missing a lot of things. Is there some guideline or something that you recommend I read up before starting any future contributions to the julia organization.

test/transforms/onehot.jl
juliohm: The change we are proposing is such that we can simply pass categorical=false to get all tests passing without change. What I am saying is that you need to add new tests instead of modifying existing ones.

juliohm:
> Also.. do beginners usually struggle so much with their first PRs sweat_smile, it seems I am missing a lot of things. Is there some guideline or something that you recommend I read up before starting any future contributions to the julia organization.

I think you start like this. At each PR you learn something new, this is the way open-source works usually. There is no well-defined recipe and each project maintainer has different standards.

src/transforms/onehot.jl
juliohm: Please remove all options as they don't help the reader of the code.

test/transforms/onehot.jl
```diff
@@ -64,10 +64,10 @@
   t = Table(; a, b)
 
   # non categorical column
-  @test_throws AssertionError apply(OneHot(:b), t)
-  @test_throws AssertionError apply(OneHot("b"), t)
+  @test_throws AssertionError apply(OneHot(:b; categorical=false), t)
+  @test_throws AssertionError apply(OneHot("b"; categorical=false), t)
```
juliohm, line 68: These changes shouldn't be necessary

test/transforms/onehot.jl
```diff
@@ -64,10 +64,10 @@
...
 
   # invalid column selection
-  @test_throws AssertionError apply(OneHot(:c), t)
-  @test_throws AssertionError apply(OneHot("c"), t)
-end
+  @test_throws AssertionError apply(OneHot(:c; categorical=false), t)
+  @test_throws AssertionError apply(OneHot("c"; categorical=false), t)
```
juliohm, line 72: same here, the assertion error will still be thrown without the option

src/transforms/onehot.jl
```diff
@@ -3,10 +3,13 @@
...
 """
-    OneHot(col)
-    
+    OneHot(col, categorical)
+
 Transforms categorical column `col` into one-hot columns of levels
 returned by the `levels` function of CategoricalArrays.jl.
+The `categorical` parameter is a bool type parameter; 
+if set to true the one-hot columns will be categorical vectors, 
+if set to false the one-hot columns will be bit vectors.
```
juliohm, line 12:
```suggestion
    OneHot(col, categorical)

Transforms categorical column `col` into one-hot columns of levels
returned by the `levels` function of CategoricalArrays.jl.
The `categorical` option can be used to convert resulting
columns to categorical arrays as opposed to boolean vectors.
```

src/transforms/onehot.jl
```diff
@@ -3,10 +3,12 @@
 # ------------------------------------------------------------------
 
 """
-    OneHot(col)
-    
+    OneHot(col, categorical)
```
juliohm, line 6:
```suggestion
    OneHot(col; categorical=true)
```

src/transforms/onehot.jl
```diff
@@ -18,9 +20,10 @@ OneHot("a")
 """
 struct OneHot{S<:ColSpec} <: StatelessFeatureTransform
   colspec::S
-  function OneHot(col::Col)
+  categorical::Bool
+  function OneHot(col::Col, categorical::Bool=true)
```
juliohm, line 24:
```suggestion
  function OneHot(col, categorical)
```

src/transforms/onehot.jl
```diff
@@ -47,7 +50,9 @@ function applyfeat(transform::OneHot, feat, prep)
   end
 
   newnms, newcols = first.(onehot), last.(onehot)
-
+  if transform.categorical == true
+    newcols = [categorical(newcol; levels=nothing, ordered=false, compress=false) for newcol in newcols]
```
juliohm, line 54:
```suggestion
  if transform.categorical
    newcols = categorical.(newcols)
```

src/transforms/onehot.jl
```diff
@@ -47,7 +50,9 @@ function applyfeat(transform::OneHot, feat, prep)
   end
 
   newnms, newcols = first.(onehot), last.(onehot)
-
+  if transform.categorical
+    newcols = categorical.(newcols)
+  end
```
juliohm, line 55:
```suggestion

  # convert to categorical arrays if necessary
  newcols = transform.categorical ? categorical.(newcols) : newcols
  
```

test/transforms/onehot.jl
juliohm, line 68:
```suggestion
  @test_throws AssertionError apply(OneHot(:b), t)
  @test_throws AssertionError apply(OneHot("b"), t)
```

test/transforms/onehot.jl
```diff
@@ -68,6 +68,6 @@
   @test_throws AssertionError apply(OneHot("b"), t)
 
   # invalid column selection
-  @test_throws AssertionError apply(OneHot(:c), t)
-  @test_throws AssertionError apply(OneHot("c"), t)
-end
+  @test_throws AssertionError apply(OneHot(:c; categorical=false), t)
+  @test_throws AssertionError apply(OneHot("c"; categorical=false), t)
```
juliohm, line 72:
```suggestion
  @test_throws AssertionError apply(OneHot(:c), t)
  @test_throws AssertionError apply(OneHot("c"), t)
```

juliohm:
@rajatrc1705 fixed a couple of issues with the proposed changes. Now we are missing tests for the new default with categorical=true. Can you please add new tests to make sure that the option is working as expected?

rajatrc1705 (commented):
The changes in this code actually break the OneHot function usage. The new code works only if both parameters are explicitly mentioned.
It doesn't work in the following case:
`x = OneHot(":b")`

image

so the previous code changes were necessary in my opinion.

rajatrc1705:
> @rajatrc1705 fixed a couple of issues with the proposed changes. Now we are missing tests for the new default with categorical=true. Can you please add new tests to make sure that the option is working as expected?

Yes I will write the new tests and commit the same here.

juliohm:
The changes here need to be fixed then because old code should work without
passing two options.

Em seg., 2 de jan. de 2023 15:40, rajatrc1705 ***@***.***>
escreveu:

rajatrc1705:
> 

Yes, for the code which I wrote, OneHot function was working without having to pass two parameters. I tested it with a few examples also. 

`b=OneHot("a")`   resulted in categorical being true

image

juliohm:
Please share text in comments instead of screenshots. It is very hard to search/copy/paste that way.

Also, try to start discussing what is needed moving forward to make things work instead of discussing the old version that wasn't ideal either.

src/transforms/onehot.jl
```diff
@@ -18,9 +20,10 @@ OneHot("a")
...
 struct OneHot{S<:ColSpec} <: StatelessFeatureTransform
   colspec::S
-  function OneHot(col::Col)
+  categorical::Bool
+  function OneHot(col, categorical)
     cs = colspec([col])
-    new{typeof(cs)}(cs)
+    new{typeof(cs)}(cs, categorical)
   end
 end
```
eliascarv, line 28: Add kwarg constructor:

```julia
struct OneHot{S<:ColSpec} <: StatelessFeatureTransform 
  colspec::S
  categorical::Bool
  function OneHot(col, categorical)
    cs = colspec([col])
    new{typeof(cs)}(cs, categorical) 
   end
 end

OneHot(col; categorical=true) = OneHot(col, categorical)

**Comment by rajatrc1705** [comment]
> Please share text in comments instead of screenshots. It is very hard to search/copy/paste that way.
> 
> Also, try to start discussing what is needed moving forward to make things work instead of discussing the old version that wasn't ideal either.

The fix shared by @eliascarv works perfectly well!
The tests for the case when categorical = true remain.

- The onehot columns for both `categorical=false` and `categorical=true` will be same, only type checking must be carried out in the new tests
  eg: `typeof(column) == typeof(categorical(Bool[]))`
- The `revertfeat` function for OneHot should be able to revert categorical arrays back to their original form, so modifications needed in the `revertfeat` function

**Comment by juliohm** [comment]
> The fix shared by @eliascarv works perfectly well!
> The tests for the case when categorical = true remain.

Looking forward to the new commits.

You can press "." on your keyboard at this page and GitHub will open the web editor where you can update the files accordingly.

**Comment by juliohm** [comment]
Thanks @eliascarv for taking the lead. We are only missing tests now with the new default.

File: src/transforms/onehot.jl [hunk 18]
```diff
@@ -48,6 +54,9 @@ function applyfeat(transform::OneHot, feat, prep)
 
   newnms, newcols = first.(onehot), last.(onehot)
 
+  # convert to categorical arrays if necessary
+  newcols = transform.categorical ? categorical.(newcols) : newcols
```
eliascarv, line 58: @juliohm, I think we should also set the levels:
```julia
categorical.(newcols, levels=[true, false])
```
For cases where the results are all `true` or all `false`

juliohm: Very good point @eliascarv , the corner case is important.

src/transforms/onehot.jl
```diff
@@ -65,7 +74,7 @@ function revertfeat(::OneHot, newfeat, fcache)
   
   oname, inds, levels, ordered = fcache
   x = map(zip(columns[inds]...)) do row
-    levels[findfirst(row)]
+    levels[findfirst(==(true), row)]
```
eliascarv, line 77: @juliohm, this code looks redundant, but it's necessary when `categorical=true`.
See:
```julia
julia> using CategoricalArrays

julia> x = Bool[0, 0, 0, 1, 1, 0]
6-element Vector{Bool}:
 0
 0
 0
 1
 1
 0

julia> findfirst(x)
4

julia> findfirst(categorical(x))
ERROR: TypeError: non-boolean (CategoricalValue{Bool, UInt32}) used in boolean context
...

**Review by juliohm — COMMENTED** [review]
Great work

File: src/transforms/onehot.jl [hunk 20]
```diff
@@ -48,6 +54,9 @@ function applyfeat(transform::OneHot, feat, prep)
 
   newnms, newcols = first.(onehot), last.(onehot)
 
+  # convert to categorical arrays if necessary
+  newcols = transform.categorical ? categorical.(newcols, levels=[true, false]) : newcols
```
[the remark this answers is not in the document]
eliascarv: Changed

juliohm (approved):

src/transforms/onehot.jl
```diff
@@ -3,34 +3,40 @@
...
 OneHot(1)
 OneHot(:a)
 OneHot("a")
+OneHot("a", categorical=false)
 ```
 """
 struct OneHot{S<:ColSpec} <: StatelessFeatureTransform
   colspec::S
-  function OneHot(col::Col)
+  categorical::Bool
```
juliohm, line 24: I wonder if we should abbreviate the field to something short like `categ`. We do similar abbreviations in `Quantile(dist=Normal())` for example to avoid typing `distribution=Normal()`.

test/transforms/onehot.jl
```diff
@@ -4,6 +4,31 @@
   c = categorical([3, 2, 2, 1, 1, 3])
   t = Table(; a, b, c)
 
+  T = OneHot(1; categorical=true)
+  n, c = apply(T, t)
+  @test Tables.columnnames(n) == (:a_false, :a_true, :b, :c)
+  @test n.a_false == categorical(Bool[1, 0, 0, 1, 0, 0])
```
rajatrc1705, line 10: Don't we need an additional type checking for the `categorical=true` case
like this:

`@test typeof(n.a_false) == typeof(categorical(Bool[]))`

cause it tests the major differentiating change of the new feature we are adding.

juliohm (approved):

juliohm:
Thanks @eliascarv ! 💯

juliohm:
Thanks @rajatrc1705 ! 💯

PR #162, opened 2023-01-14
New Assertion API

### LINKED ISSUE #93: More general assertion API
Currently we provide a few basic checks as assertion functions that can be used before the transform is applied. For example, some transforms like EigenAnalysis only make sense with continuous variables. We need to refactor this API and consider assertions that apply to a specific ColSpec and assertions that hold state.

In particular, we could return a list of structs in the `assertions` function that represent different types of assertion. For example, we could consider something like:

```julia
"""
    SciTypeAssertion{T}(colspec)

Asserts that the columns in the `colspec` have a scientific type `T`.
"""
struct SciTypeAssertion{T,S}
  colspec::S
end
```

We could then introduce an API `apply(assertion, table)` to throw an error whenever the checks fail. The name of the function could also be something else to avoid confusion with the `apply` of transforms.

eliascarv:
closes #93

### DIFF
--- a/src/TableTransforms.jl
+++ b/src/TableTransforms.jl
@@ -26,10 +26,10 @@ import TransformsBase: assertions, isrevertible, preprocess
 import TransformsBase: apply, revert, reapply
 import TransformsBase: Identity, →
 
-include("tabletraits.jl")
-include("assertions.jl")
-include("distributions.jl")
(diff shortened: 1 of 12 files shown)


juliohm (approved):
Great cleanup 💯

PR #182, opened 2023-08-21
add docs for custom transforms

### LINKED ISSUE #175: Implementing the TableTransforms.jl transformer interface in a different package
I'm working on a small package for oversampling input tables. I thought it would be nice if I make it possible to use the methods in the package using a similar API to the one here but I couldn't find the relevant section to help me do that in the docs.

EssamWisam:
@juliohm please take a look at the docs.

Fix #175


docs/make.jl
```diff
@@ -27,7 +27,7 @@ makedocs(;
       asset("https://fonts.googleapis.com/css?family=Montserrat|Source+Code+Pro&display=swap", class=:css)
     ]
   ),
-  pages=["Home" => "index.md", "Transforms" => "transforms.md", "Related" => "related.md"]
+  pages=["Home" => "index.md", "Transforms" => "transforms.md", "Custom Transforms"=>"custom.md" , "Related" => "related.md"]
```
juliohm, line 30:
```suggestion
  pages=["Home" => "index.md", "Transforms" => "transforms.md", "Developer guide"=>"devguide.md" , "Related" => "related.md"]
```

docs/src/custom.md
```diff
@@ -0,0 +1,126 @@
+# Custom Transforms
+A short guide for extending the interface with your own custom transform.
```
juliohm, line 2:
```suggestion
# Developer guide

A short guide for extending the interface as a developer.
```

docs/src/custom.md
```diff
@@ -0,0 +1,126 @@
...
+
+## Motivation
+
+TableTransforms currently supports over 25 different transforms that cover a wide variety of use cases ranging from ordinary table operations to complex statistical transformations and that all can be arbitrarily composed with one another through elegant syntax. It is easy to leverage all of those benefits as a developer of a new transforms that operates on tables so that it perfectly integrates with the rest of the TableTransforms ecosystem.
```
juliohm, line 6:
```suggestion
TableTransforms.jl currently supports over 25 different transforms that cover a wide variety of
use cases ranging from ordinary table operations to complex statistical transformations, which
can be arbitrarily composed with one another through elegant syntax. It is easy to leverage all
this functionality as a developer of new transforms, and this is the motivation of this guide.
```

docs/src/custom.md
```diff
@@ -0,0 +1,126 @@
...
+##  Custom Transform Basics
+All the transforms in this package implement the transforms interface defined in the `TransformsBase` package so this is really the only dependency you need. The interface assumes the following about your implementation:
+
+- Your transform operates on a single table 
+- Your transform may be associated with some state that if computed while applying it for the first time and is then cached can help later reapply the transform on another table without recomputing the state
+- Your transform may be revertible, meaning that a transformed table can be brought back to its original form, and it may need to use the cache for that
+- Your transform may be invertible in the mathematical sense
+
+!!! note "Revertible does not imply invertible"
+    Reverting assumes that the transform has been applied already and reverts it. Meanwhile, invertible implies that there is a one-to-one mapping between the input and output tables so a table can be inverted to a corresponding input to the transform even if it was not transformed a priori.
```
juliohm, line 17:
```suggestion
##  Basic assumptions

All the transforms in this package implement the transforms interface defined in the TransformsBase.jl
package so this is really the only dependency needed. The interface assumes the following about new transforms:

- The transform operates on a single table
- The transform may be associated with some state that if computed while applying it for the first time and is then cached can help later reapply the transform on another table without recomputing the state
- The transform may be revertible, meaning that a transformed table can be brought back to its original form, and it may need to use the cache for that
- Your transform may be invertible in the mathematical sense

!!! note "Revertible does not imply invertible"
    Reverting assumes that the transform has been applied already and reverts it. Meanwhile, invertible implies that there is a one-to-one mapping between the input and output tables so a table can be inverted to a corresponding input to the transform even if it was not transformed a priori.
```

docs/src/custom.md
```diff
@@ -0,0 +1,126 @@
...
+struct Standardizer <: TransformsBase.Transform
+    center::Bool
+    scale::Bool
+end
+```
+You may implement keyword instructors as needed if some of the hyperparameters are optional. Thus, we follow up with:
+```julia
+Standardize(;center::Bool=true, scale::Bool=true) = Standardize(center, scale)
+```
+
```
juliohm, line 35:
```suggestion
## Defining a new transform

In the following we shall demonstrate the steps to define a new transform.

### 1. Declare a new type for your transform

The type should subtype `TransformsBase.Transform` and it should have a named field for each input needed to apply the transform besides the input table. For instance, if you want to call your transform `Standardize` and it takes two boolean inputs `center` and `scale`, then you should declare:

```julia
struct Standardize <: TransformsBase.Transform
    center::Bool
    scale::Bool
end
```

You may implement keyword constructors as needed if some of the hyperparameters are optional.
Thus, we follow up with:

```julia
Standardize(;center::Bool=true, scale::Bool=true) = Standardize(center, scale)
```

```

File: docs/src/custom.md [hunk 6]
```diff
@@ -0,0 +1,126 @@
...
+### 2. Implement the `apply` method for your transform
+The `apply` method takes an object of your transform type and a table and returns a new table and cache. Suppose that the `Standardize` transform should zero-mean each column if `center` is true and scale each column to unit variance if `scale` is true, then the `apply` method should be implemented as follows:
```
juliohm, line 37:
```suggestion
### 2. Implement the `apply` method for your transform

The `apply` method takes an instance of your transform type and a table and returns a new table and cache. Suppose that the `Standardize` transform should zero-mean each column if `center` is true and scale each column to unit variance if `scale` is true, then the `apply` method should be implemented as follows:

```

docs/src/custom.md
```diff
@@ -0,0 +1,126 @@
...
+    # standardize the data
+    Xm = (Xm .- μ) ./ σ
+    # convert back to table
+    Xm = X |> Tables.materializer(Xm)
+    # return the table and cache that may help reapply or revert later
+    return Xm, (μ, σ)
+end
+```
+
+That's it really! Your transform now behaves like any table transform. For instance, we can do
```
juliohm, line 57:
```suggestion
That's it really! Your transform now behaves like any table transform. For instance, we can do

```

docs/src/custom.md
```diff
@@ -0,0 +1,126 @@
...
+using TableTransforms
+
+X = (A=[1, 2, 3], B=[4, 5, 6])
+Xt = X |> Standardize() |> Identity() |> Select([:A])
+```
+
+It holds, however, that in case your transform can be reapplied, is revertible, or is invertible then you should continue implementing the interface to support such functionality.
+
+### 3. Optionally implement `reapply`
+We need this in case of the `Standarize` transform because after computing the mean and std for some train table we may want to apply the transform directly given a test table. Hence, we implement `reapply` which has the same signature as apply but it takes an extra argument for the cache and doesn't return it.
```
juliohm, line 68:
```suggestion
### 3. Optionally implement `reapply`

We need this in case of the `Standarize` transform because after computing the mean and std for some train table we may want to apply the transform directly given a test table. Hence, we implement `reapply` which has the same signature as apply but it takes an extra argument for the cache and doesn't return it.

```

docs/src/custom.md
```diff
@@ -0,0 +1,126 @@
...
+    μ, σ = cache
+    # standardize the data
+    Xm = (Xm .- μ) ./ σ
+    # convert back to table
+    Xm = X |> Tables.materializer(Xm)
+    return Xm
+end
+```
+
+
```
juliohm, line 84:
```suggestion
```

docs/src/custom.md
```diff
@@ -0,0 +1,126 @@
...
+    μ, σ = cache
+    # revert the transform
+    Xm = Xm .* σ .+ μ
+    # convert back to table
+    Xm = X |> Tables.materializer(Xm)
+    return Xm
+end
+```
+
+### 5. Optionally specify that your transform is invertible and implement `Base.inv`
```
juliohm, line 110:
```suggestion
### 5. Optionally specify that your transform is invertible and implement `Base.inv`

```

docs/src/custom.md
```diff
@@ -0,0 +1,126 @@
...
+Similar to reversibility, falls back to false by default. We can write that explicitly here since `Standardize` has no inverse if we are given nothing except for the table.
+
+```julia
+TransformsBase.isinvertible(::Type{Standardize}) = false
+```
+
+If an arbitrary transform `T` is invertible we can rather specify that as true and follow up by implementing `Base.inv(::T)` which would be expected to return an instance of the inverse transform. For instance, for an identity transform we can do
```
juliohm, line 117:
```suggestion
If an arbitrary transform `T` is invertible we can rather specify that as true and follow up by implementing `Base.inv(::T)` which would be expected to return an instance of the inverse transform. For instance, for an identity transform we can do

```

juliohm (changes requested):
Awesome docs @EssamWisam 👏🏽 I added a few minor suggestions.

After you review the suggestions, can you please rename the file to devguide.jl and use a maximum number of characters per line to facilitate reviews in the future? Git diffs are terrible with long lines.

docs/src/custom.md
EssamWisam: I think this commit got messed up. Will fix it.

juliohm (approved):

PR #191, opened 2023-09-03
Fix custom transforms docs example

EssamWisam:
Last time when I worked on the docs, I misused `Tables.materialize` without noticing. When I ran the example, for some reason, `Tables.jl` did not return an error and instead returned the original table and I didn't notice that anything is wrong because I didn't focus on the specific numbers.

This issue is solved in the PR.


docs/make.jl
```diff
@@ -27,6 +27,7 @@ makedocs(;
       asset("https://fonts.googleapis.com/css?family=Montserrat|Source+Code+Pro&display=swap", class=:css)
     ]
   ),
+
```
juliohm, line 30:
```suggestion
```

Issue #193, opened 2023-09-04
DropExtrema


juliohm:
The `DropExtrema(low=0.01, high=0.99)` transform drops extreme values in both lower and upper tails.

Jay-sanjay:
sir , where is the function DropExtrema defined I am not getting it in the TableTransform.jl

juliohm:
The issue is about adding the transform, it doesn't exist yet.

Em sex., 8 de set. de 2023 04:26, Jay-sanjay ***@***.***>
escreveu:

juliohm:
The `DropExtrema` transform is a transform that computes `a = quantile(column, low)` and `b = quantile(column, high)`, and then discards all rows for which the values of the `column` lie outside the interval `[a,b]`.

Jay-sanjay:
Sir , this is the link for the PR
https://github.com/JuliaML/TableTransforms.jl/pull/197

juliohm:
Fixed in the master branch.

PR #197, opened 2023-09-08
Added the DropExtrema function

Jay-sanjay:
Issue  #193 fixing
Added the actual function for DropExtrema ,


src/TableTransforms.jl
```diff
@@ -57,6 +57,7 @@ export
   Replace,
   Coalesce,
   Coerce,
+  dropExtrema,
```
juliohm, line 60: Notice that we have a naming convention that needs to be followed. The transforms always start with a capital letter.

src/transforms/dropExtrema.jl
```diff
@@ -0,0 +1,27 @@
+# Licensed under the MIT License. See LICENSE in the project root.
+# ------------------------------------------------------------------
+
+"""
+        DropExtrema(low, high)
```
juliohm, line 5: Try to follow the docstring conventions `Scale` transform as it also has a low and a high.

src/transforms/dropExtrema.jl
```diff
@@ -0,0 +1,27 @@
...
+    
+    Applies the DropExtrema transform to the specified column `col` with interval [low, high].
+    
+    # Examples
+    ```julia
+    DropExtrema(0.01, 0.99)
+    DropExtrema(:column_name, 1.0, 10.0)
+    
+"""
+function drop_extrema(df::DataFrame, column_name::Symbol, low::Float64, high::Float64)
```
juliohm, line 20: You need to implement the same functions that are implemented for other transforms. See the developer guide for more information: https://juliaml.github.io/TableTransforms.jl/stable/devguide/

juliohm (changes requested):

src/TableTransforms.jl
Jay-sanjay: I will update it sir , thanks for pointing out

src/transforms/dropExtrema.jl
Jay-sanjay: I will update it sir similar to what is done in other transform methods , thanks for pointing out.

src/transforms/dropExtrema.jl
Jay-sanjay: yeah sir I will check it

Jay-sanjay:
@juliohm  sir I am now working on the testCases part, which is according to me the potential reason for failing checks
Sir can you please tell what command to run in order to run the complete project locally in my pc.

src/transforms/dropExtrema.jl
```diff
@@ -0,0 +1,35 @@
...
+    col::Symbol
+end
+
+DropExtrema(low::Real, high::Real) = DropExtrema(low, high, :)
+DropExtrema(col::Symbol, low::Real, high::Real) = DropExtrema(low, high, col)
+
+function TransformsBase.apply(transform::DropExtrema, X)
+    a = quantile(X[!, transform.col], transform.low)
+    b = quantile(X[!, transform.col], transform.high)
+    filtered_rows = filter(row -> a <= row[!, transform.col] <= b, 1:nrow(X))
```
juliohm, line 33: take a look at the other transforms, we don't use DataFrame syntax, we need to use Tables.jl as an interface.

src/transforms/dropExtrema.jl
```diff
@@ -0,0 +1,35 @@
...
+# Examples
+```julia
+DropExtrema(0.01, 0.99)
+DropExtrema(:column_name, 1.0, 10.0)
+```
+"""
+struct DropExtrema{T<:Real} <: TransformsBase.Transform
+    low::T
+    high::T
+    col::Symbol
```
juliohm, line 24: all other files use 2 whitespaces for indentation

src/transforms/dropExtrema.jl
```diff
@@ -0,0 +1,35 @@
...
+end
+
+DropExtrema(low::Real, high::Real) = DropExtrema(low, high, :)
+DropExtrema(col::Symbol, low::Real, high::Real) = DropExtrema(low, high, col)
```
juliohm, line 28: these options should be kwargs instead of args

test/transforms.jl
```diff
@@ -24,6 +24,7 @@ transformfiles = [
   "coltable.jl",
   "sequential.jl",
   "parallel.jl"
+  "dropExtrema.jl"
```
juliohm, line 27: the name of source files should be lowercase

juliohm (changes requested):
still lots to go. try to read the Julia documentation, the Tables.jl documentation and then come back here to contribute. In particular, try to ask in Zulip or other forums these basic user questions about tests, etc. people can help there as well.

here we can give you support about specifics of TableTransforms.jl

juliohm:
Implemented in #201

PR #224, opened 2023-11-06
Fix #223 + Some refactors in `tablerows` and `Parallel`

### LINKED ISSUE #223: Pipes with parallel branches can return invalid tables
```julia
using TableTransforms

t = (a=1:4, b=5:8)

left = Select(:a) → Filter(row -> row.a < 4)
right = Select(:b) → Filter(row -> row.b < 7)
transform = left ⊔ right;
```

```julia
julia> transform(t)
(a = [1, 2, 3], b = [5, 6])
```

This should probably throw an error. But even if it did throw an error, it would illustrate a degree of freedom in the API design that allows the user to shoot themselves in the foot.

(The API design solution in my mind is to remove table queries from TableTransforms.jl and stick to ML and statistics transforms that preserve the number of rows of the input table.)

eliascarv:
fix #223

### DIFF
--- a/src/transforms/parallel.jl
+++ b/src/transforms/parallel.jl
@@ -53,6 +53,9 @@ function apply(p::ParallelTableTransform, table)
   feats = first.(splits)
   metas = last.(splits)
 
+  # check the number of rows of generated tables
+  @assert allequal(_nrows(f) for f in feats) "parallel branches must produce the same number of rows"
+
   # table with concatenated features
   newfeat = tablehcat(feats)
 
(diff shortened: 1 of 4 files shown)


juliohm (approved):

src/transforms/parallel.jl
```diff
@@ -53,6 +53,9 @@ function apply(p::ParallelTableTransform, table)
   feats = first.(splits)
   metas = last.(splits)
 
+  # check the number of rows of generated tables
+  @assert allequal(_nrows(f) for f in feats) "parallel branches must produce the same number of rows"
```
CameronBieganek, line 57: From the `@assert` docstring:

> An assert might be disabled at various optimization levels.

So, `@assert` should only be used for debugging. For actual interface behavior you should `throw` an exception.

PR #226, opened 2023-11-06
Replace the `assert` macro with the new `_assert` utility function

eliascarv:
Motivation: `@assert` docstring:
```
help?> @assert
  @assert cond [text]

  Throw an AssertionError if cond is false. Preferred syntax for writing
  assertions. Message text is optionally displayed upon assertion failure.

  │ Warning
  │
  │  An assert might be disabled at various optimization levels. Assert
  │  should therefore only be used as a debugging tool and not used for
  │  authentication verification (e.g., verifying passwords), nor should
  │  side effects needed for the function to work correctly be used
  │  inside of asserts.
```

### DIFF
--- a/src/assertions.jl
+++ b/src/assertions.jl
@@ -22,6 +22,12 @@ function (assertion::SciTypeAssertion{T})(table) where {T}
 
   for nm in snames
     x = Tables.getcolumn(cols, nm)
-    @assert elscitype(x) <: T "the elements of the column '$nm' are not of scientific type $T"
+    _assert(elscitype(x) <: T, "the elements of the column '$nm' are not of scientific type $T")
   end
 end
+
+#-------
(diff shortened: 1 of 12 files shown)


juliohm (approved):

CameronBieganek:
Nitpick:
This is better than using `@assert`, but it is more idiomatic to throw an `ArgumentError` than to throw an `AssertionError`.

juliohm:
ArgumentError is for when the user misuses a function right? In the case of
the parallel branches we thought that the user can safely create pipelines
and that the error will only be encountered at runtime depending on the
tables that are inserted.

We can reassess the type of error in the future.

Em ter., 7 de nov. de 2023 11:58, Cameron Bieganek ***@***.***>
escreveu:

PR #288, opened 2024-06-24
Use `Distributed.pmap` in `ParallelTableTransform` and `ColwiseFeatureTransform`

### LINKED ISSUE #286: Use `Distributed` processes in `ParallelTableTransforms`
Threads are useful when a simple operation (ex: +, -, ...) is applied to multiple data. Transform pipelines are much more sophisticated than that in industrial applications, and call for distributed processes instead.

eliascarv:
closes #286


juliohm:
We need to benchmark the change, compare it with the previous code. Make sure that `pmap` is using the `CachePool` if needed. I know it is the default in latest Julia versions, please double check if Julia v1.9 has it as the default.

juliohm:
Additionally, we use `tcollect` in ColwiseTransform, and should probably use a `pmap` too. If that is indeed the case, please update the code, drop Transducers.jl from the list of deps, and update the documentation accordingly.

juliohm (approved):

Issue #290, opened 2024-09-17
Map accepting row functions


davibarreira:
It seems that the `Map` transform does not accept row functions, for example:
```julia
Map(row->row.a + row.b => :NewColumnName)
Map(row->sum(row) => :NewColumnName)
```
Is this by design? I've taken a look at the implementation of how `Map` is applied, and I noticed that the data is first transformed into column form, and then a generator is produced. Isn't the `Map` transform always applied row-wise? Is there a reason to parsing the table column-wise instead?

Using the `ColumnSelectors.AllSelector` type, one can place an `if` and apply the function row-wise, without modifying much of the current implementation. But this is not a perfect solution, as it produces errors if the user actually pass an AllSelector and writes the function with `(a,b,c,d)` containing all columns.

juliohm:
Good question. We are effectively mapping a function to all rows of the table, so we could potentially consider the proposed syntax. However, not all `row` objects support the syntax `row.a` so we need to make sure that the table type is wrapped into a new table type with this guarantee.

@eliascarv do you have any comment regarding this proposal? At a quick glance, it can drastically simplify the `Map` syntax.

eliascarv:
@davibarreira, after discussion, perhaps the best decision would be to change the current syntax to this:
```julia
Map(fun, ...)
Map(fun => target, ....)
```
Where `fun` is a function that receives a `row` and returns a value: `fun(row) -> value`.

For example, the following code:
```julia
Map(:a => sin, "b" => cos => :cos_b)
```
Would be written like this:
```julia
Map((row -> sin(row.a)), (row -> cos(row.b)) => :cos_b)
```
Can you do a PR to investigate this please?

juliohm:
To clarify, this PR would be investigative. We can compare the pros and cons of the two approaches and decide later if the breaking change is worth.

davibarreira:
I've implemented the solution I mentioned and submitted the PR. You guys can check if you think it properly addresses the issue, or if it is better to keep as it is now.

juliohm:
Thank you @davibarreira for experimenting with this idea. Closing the issue.

PR #291, opened 2024-09-17
Implementing row function syntax to Map transform

davibarreira:
Addressing #290.

Enables row functions such as:
```julia

Map(row->row.a + row.b => :NewColumnName)
Map(row->sum(row) => :NewColumnName)
```

This PR has the following limitation:
```julia
a = [4, 7, 8, 5, 8, 1]
b = [1, 9, 1, 7, 9, 4]
c = [2, 8, 6, 3, 2, 2]
d = [7, 5, 9, 5, 3, 4]
t = Table(; a, b, c, d)

t |> Map(row->row.a) # This works!

t |> Map((a,b,c,d)->a) # This fails!
```

### DIFF
--- a/src/transforms/map.jl
+++ b/src/transforms/map.jl
@@ -48,11 +48,14 @@ Map() = throw(ArgumentError("cannot create Map transform without arguments"))
 const TargetName = Union{Symbol,AbstractString}
 const PairWithTarget = Pair{<:Any,<:Pair{<:Function,<:TargetName}}
 const PairWithoutTarget = Pair{<:Any,<:Function}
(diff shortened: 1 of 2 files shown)


eliascarv:
I think I may have expressed myself poorly, @davibarreira, I'm sorry.
But the idea was to stop supporting the current syntax and only support this syntax:
```julia
Map(fun_1, fun_2 => target_2, ...)
```
Where `fun` is a function that takes a `row` and returns a value (`fun(row) -> value`) and `target` is optional.

davibarreira:
No problem. I thought preserving the current syntax would be desired since it would avoid a breaking version.

PR #292, opened 2024-09-18
Refactor Map with row function syntax #291

davibarreira:
Addressing https://github.com/JuliaML/TableTransforms.jl/issues/290.

Enables row functions such as:

```julia

Map((row->row.a + row.b) => :NewColumnName)
Map(row->sum(row) => :NewColumnName)
```

This is just an exploration, so I did not fix the test for `show`.


davibarreira:
I haven't checked. I'll do some benchmarks to see if there is improvement.

juliohm:
In any case I think the syntax is limited compared to what we can do today.
Should we just close it? The performance benefit may not pay off in this
case.

Em qui., 19 de set. de 2024, 13:27, Davi Sales Barreira <
***@***.***> escreveu:

PR #298, opened 2025-03-27
Add KMedoids

juliohm:


src/transforms/kmedoids.jl
```diff
@@ -0,0 +1,145 @@
...
+
+    X = Tables.subset(table, inds)
+
+    j = _medoid(td, X)
+
+    @inbounds medoids[k] = inds[j]
+  end
+end
+
+function _nrow(table)
```
eliascarv, line 134: This utility function already exists in `tablerows.jl` file.

juliohm: Thank you! Will use it instead 👍🏽

src/transforms/kmedoids.jl
```diff
@@ -0,0 +1,138 @@
...
+  k::Int
+  tol::Float64
+  maxiter::Int
+  weights::W
+  rng::RNG
+end
+
+function KMedoids(k; tol=1e-4, maxiter=10, weights=nothing, rng=Random.default_rng())
+  # sanity checks
+  @assert k > 0 "number of clusters must be positive"
```
eliascarv, line 45: Please, replace the `@assert` macro by the `_assert` helper function.

eliascarv:
I think the only point that we don't have a simple way to resolve would be the type of `dists = fill(Inf, nobs)`, as we don't have a practical way of knowing what the type returned by the `parwise` of the columns would be.

So I think starting it as `Float64` is the simplest and safest solution, just like it is in the PR.

eliascarv (approved):

juliohm:
Perhaps we can evaluate the distance with a single row to get the type.
Will try when I'm back to the computer.

Em qui., 27 de mar. de 2025, 19:10, Elias Carvalho ***@***.***>
escreveu:

juliohm:
There is still an issue regarding unitful columns with different units, but that should be addressed in TableDistances.jl. If distances are computed with different units, and these units can't be added, we will get an error. Perhaps TableDistances.jl should remove the units to make it work, or warn users that columns must have compatible units.

juliohm:
I am going to merge assuming that the last review was an approval with comment 👍🏽

eliascarv:
Amazing PR @juliohm!

PR #299, opened 2025-07-17
Refactor TableSelection

juliohm:
This PR refactors the internal `TableSelection` type to avoid runtime checks in the wrong places. This type is intended for developers of TableTransforms.jl so the only check that should occur about the columns names is in the constructor. All further calls to Tables.getcolumn should assume the existence of columns for performance.

I also refactored the field type to a static tuple to improve the performance in downstream algorithm given the number of columns at compile time. It is also more consistent with other table types that return tuples in Tables.columnnames.

@eliascarv is there any particular reason we used Vector{Symbol} for the names or are we ok replacing it with tuple?

### DIFF
--- a/src/tableselection.jl
+++ b/src/tableselection.jl
@@ -10,15 +10,12 @@ Stores a sub-`table` with given column `names`.
 struct TableSelection{T,N}
   table::T
   names::NTuple{N,Symbol}
-  oinds::NTuple{N,Int}
 end
 
 function TableSelection(table, names)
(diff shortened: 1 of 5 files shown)


eliascarv:
No particular reason. When the tests pass I will approve.

eliascarv (approved):

juliohm:
I will do an extra refactor to remove the `oinds` field. It is not needed either.

PR #302, opened 2025-08-06
Return target columns in Map

juliohm:
@eliascarv I remember we opted to return all columns of the original table to mimic DataFrames.jl's transform behavior. Is there any other reason? A more ergonomic behavior is proposed in this PR that only returns the new (target) columns in the result.


eliascarv:
@juliohm There was no other reason. I remember that mimicking the behavior of DataFrames.jl was the only reason.

eliascarv (approved):

PR #327, opened 2026-07-18
Add nmax option to KMedoids

juliohm:
To handle datasets that don't fit in memory.


juliohm:
Failing test is random due to GitHub infrastructure and running times. Merging it.

